PR #45, opened 2020-08-14
Km37 keyring

ivaruf:
## Description
Store user password to aws saml assertion so that the user does not have to re-type it every time they do a call that uses saml.

## Motivation and Context
Better user experience and a nice task to get started that is not very inter-connected with the whole application and could be solved in isolation

## How Has This Been Tested?
Tested on Linux with okctl, which resulted in a gnome-keyring that is reused, a couple of tests also on Store / Fetch in memory keyring

## Types of changes

## Checklist:


ivaruf:
Needs cleanup and some linting - as well as description

PR #55, opened 2020-08-21
Change method to install goreleaser, no longer able to get from "go get"

ivaruf:
For details see: https://github.com/goreleaser/goreleaser/issues/1405

## Description
For the issue of using "go get" with goreleaser, make release-local has been changed to download goreleaser each time it runs (with curl, and not go get). See here for more details: https://github.com/goreleaser/goreleaser/issues/126

## Motivation and Context
Make release-local failed in linux because of an issue with "go get"

## How Has This Been Tested?
Tested locally

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ GOCOV          := $(GOBIN)/gocov
 RICHGO         := $(GOBIN)/richgo
 MAKEDOC        := $(GOBIN)/makedoc
 STATIK         := $(GOBIN)/statik
-GORELEASER     := goreleaser
+GORELEASER     := bin/goreleaser/v.0.142.0/$(OS)/goreleaser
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -97,7 +98,7 @@ $(STATIK):
 	$(GO) get -u github.com/rakyll/statik
 
 $(GORELEASER):
```
paulbes, line 100: We can drop this block and rather move `bin/goreleaser/v0.142.0/$(OS)/goreleaser` up to the `GORELEASER := bin/goreleaser/v0.142.0/$(OS)/goreleaser` statement.

paulbes (changes requested):
Love the approach 👍 one small change is required though

Makefile
ivaruf: Done :+1:

PR #85, opened 2020-09-23
KM68 service quotas

ivaruf:
Check quotas for services on AWS account before cluster creation

## Description
We were hitting errors on quotas for VPCs, Addresses and Internet gateways during cluster creation process. To remedy this we check that the resources we need are available before cluster creation starts.

## Motivation and Context
Better user experience to be told what will go wrong before it happens 

## How Has This Been Tested?
Tested manually

## Types of changes

## Checklist:


pkg/servicequota/servicequota.go
```diff
@@ -0,0 +1,89 @@
...
+func getLengthOf(result map[string]interface{}, path string) (int, error) {
+	if result[path] == nil {
+		return 0, nil
+	}
+
+	return len(result[path].([]interface{})), nil
+}
+
+func getQuotaNumericValue(quota *servicequotas.GetServiceQuotaOutput) (int, error) {
+	return strconv.Atoi(getValueFromHjson(quota.String(), "Quota.Value"))
```
paulbes, line 88: Rather than fetching the value this way, we can derefence the value in the returned output: https://docs.aws.amazon.com/sdk-for-go/api/service/servicequotas/#ServiceQuota.

e.g.:

```go
output := &ServiceQuota{}
value := *output.Value
```

pkg/servicequota/servicequota.go
```diff
@@ -0,0 +1,89 @@
...
+type Usage interface {
+	Count() (int, error)
+	Quota() (int, error)
+	Required() int
+	CheckAvailability() error
+	Description() string
+}
+
+// CheckQuotas check if we have enough resources for given cloud provider
+func CheckQuotas(provider v1alpha1.CloudProvider) error {
```
paulbes, line 24: Here we should accept a list of the `Usage` interfaces, rather than the `v1alpha1.CloudProvider`, e.g.:

```go
func CheckQuotas(checks []Usage) error { ... }
```

We can then iterate over the checks to determine if they are good or not.

pkg/servicequota/servicequota.go
```diff
@@ -0,0 +1,89 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/service/servicequotas"
+	"github.com/hjson/hjson-go"
+	"github.com/oslokommune/okctl/pkg/api/okctl.io/v1alpha1"
+	"github.com/tidwall/gjson"
+)
+
+// Usage defines what we need to know about a service quota
+type Usage interface {
```
paulbes, line 15: Maybe we can simplify this interface, seems we only need `CheckAvailability`? Also, it is idiomatic to name the interface with an `er` extension (https://golang.org/doc/effective_go.html#interface-names), so maybe `Checker` is a better name than `Usage`?

pkg/servicequota/vpcusage.go
```diff
@@ -0,0 +1,71 @@
+package servicequota
```
paulbes, line 1: We should create a function for creating a new quota checker, e.g.:

```go
func NewVpcChecker(provider v1alpha1.CloudProvider) *VpcChecker {
  return &VpcChecker{
    provider: provider,
  }
}
```

Or something to this effect.

pkg/servicequota/vpcusage.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go/service/servicequotas"
+	"github.com/logrusorgru/aurora"
+	"github.com/oslokommune/okctl/pkg/api/okctl.io/v1alpha1"
+)
+
+type vpcusage struct {
+	CloudProvider v1alpha1.CloudProvider
+}
+
+func (v vpcusage) Count() (int, error) {
```
paulbes, line 18: Prefer to operate on a pointer to the struct, otherwise, we are doing a lot of copy by value, which is good when we want to protect the internal data/state, but that is not the case for these checks.

pkg/servicequota/vpcusage.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+
+	vpcCount, err := v.Count()
+	if err != nil {
+		return err
+	}
+
+	vpcAvail := vpcQuota - vpcCount
+	fmt.Printf("You have %d VPCs, and the quota for %s is %d,\n%d are available and you need %d", vpcCount, v.CloudProvider.Region(), vpcQuota, vpcAvail, v.Required())
```
paulbes, line 59: We should not be writing directly to `os.Stdout`, which is what `fmt.Printf` does. If we want to write, we should also receive an `io.Writer`, and then use `fmt.Fprintf` and family. This is preferable, because we can then pass in `ioutil.Discard`, which simply discards what we are writing, or even `os.Stderr`.

If you are passing in an `io.Writer` to the `NewVpcCheck`, then prefer to use the one configured in okctl, e.g., `o.Err`, also prefer err, since we don't want to clutter `os.Stdout` with log messages.

cmd/okctl/create.go
```diff
@@ -255,6 +257,11 @@ and database subnets.`,
 				Message: "Are you ready to start?",
 			}
 
+			err = servicequota.CheckQuotas(o.CloudProvider)
```
paulbes, line 260: Here we should be newing up the checks, and passing them in to the checker implementation, it is easier to control the setup of the individual checkers, which ones we want to add, etc. See the comments further down the stack 👍

pkg/servicequota/eipusage.go
```diff
@@ -0,0 +1,71 @@
...
+	})
+	if err != nil {
+		return -1, fmt.Errorf("failed to get eip quota: %w", err)
+	}
+
+	return getQuotaNumericValue(quotas)
+}
+
+func (eipusage) Required() int {
+	return config.DefaultRequiredEpis
```
paulbes, line 40: We should probably inject the required number of "things" when newing up the struct.

pkg/servicequota/servicequota.go
```diff
@@ -0,0 +1,89 @@
...
+func CheckQuotas(provider v1alpha1.CloudProvider) error {
+	vpcs := vpcusage{}
+	vpcs.CloudProvider = provider
+
+	err := vpcs.CheckAvailability()
+	if err != nil {
+		return err
+	}
+
+	eips := eipusage{}
```
paulbes, line 33: We should be injecting these checks, otherwise, we are locked down to the same checks, always, see my other comments 👍

paulbes (commented):
Great start! I have some comments on various language idioms, and patterns that help us break up the code more.

pkg/config/config.go
```diff
@@ -129,6 +129,13 @@ const (
 
 	// DefaultKeyringServiceName is the name of the keyring or encrypted file used to store client secrets
 	DefaultKeyringServiceName = "okctlService"
+
+	// DefaultRequiredEpis number of elastic ips required for cluster creation
+	DefaultRequiredEpis = 3
```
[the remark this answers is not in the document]
ivaruf: Ah, good point, will have to override this for testcluster

pkg/servicequota/eipusage.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type eipusage struct {
+	CloudProvider v1alpha1.CloudProvider
+}
+
+func (e eipusage) Count() (int, error) {
+	eips, err := e.CloudProvider.EC2().DescribeAddresses(nil)
+	if err != nil {
+		return -1, fmt.Errorf("failed to get eip count: %w", err)
```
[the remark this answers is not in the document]
ivaruf: Ok, done!

pkg/servicequota/servicequota.go
ivaruf: Could you provide a more complete example?  I don't get what you are saying here at all.

paulbes: Absolutely! Does this snippet of code demonstrate it better: https://github.com/oslokommune/okctl/blob/master/pkg/route53/route53.go#L38-L58?

Basically you don't need to use the `.String()` function on the output and parse that output. You can rather dereference the field you are interested in. So, for the `ServiceQuota` struct, we have this form:

```go
type ServiceQuota struct {

    // Specifies if the quota value can be increased.
    Adjustable *bool `type:"boolean"`

    // Specifies the ErrorCode and ErrorMessage when success isn't achieved.
    ErrorReason *ErrorReason `type:"structure"`

    // Specifies if the quota is global.
    GlobalQuota *bool `type:"boolean"`

    // Identifies the unit and value of how time is measured.
    Period *QuotaPeriod `type:"structure"`

    // The Amazon Resource Name (ARN) of the service quota.
    QuotaArn *string `type:"string"`

    // The code identifier for the service quota specified.
    QuotaCode *string `min:"1" type:"string"`

    // The name identifier of the service quota.
    QuotaName *string `type:"string"`

    // Specifies the service that you want to use.
    ServiceCode *string `min:"1" type:"string"`

    // The name of the AWS service specified in the increase request.
    ServiceName *string `type:"string"`

    // The unit of measurement for the value of the service quota.
    Unit *string `type:"string"`

    // Specifies the details about the measurement.
    UsageMetric *MetricInfo `type:"structure"`

    // The value of service quota.
    Value *float64 `type:"double"`
    // contains filtered or unexported fields
}
```

If you have an instance of this type above, which you have in this receiver: `quota *servicequotas.GetServiceQuotaOutput`, you can simply dereference `Value` for example using `*quota.Quota.Value`

ivaruf: Ah ... this makes a lot of sense, so now i can get rid of the humanJson parsing bit, thanks! :)

pkg/servicequota/servicequota.go
ivaruf: Yepp, done :)

cmd/okctl/create.go
ivaruf: Done

pkg/servicequota/eipusage.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/oslokommune/okctl/pkg/config"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/servicequotas"
+	"github.com/logrusorgru/aurora"
+	"github.com/oslokommune/okctl/pkg/api/okctl.io/v1alpha1"
+)
+
+type eipusage struct {
+	CloudProvider v1alpha1.CloudProvider
```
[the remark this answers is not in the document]
ivaruf: This is on old code, but how could we call CloudProvider (and pass it to the struct) if we do not export it?

pkg/servicequota/vpcusage.go
ivaruf: Done :)

pkg/servicequota/servicequota.go
ivaruf: Agree, done :+1:

pkg/servicequota/vpcusage.go
ivaruf: Done :)

pkg/servicequota/vpcusage.go
ivaruf: Done :)

pkg/servicequota/servicequota.go
ivaruf: Done :+1:

cmd/okctl/create.go
```diff
@@ -250,6 +252,13 @@ and database subnets.`,
 				return formatErr(err)
 			}
 
+			var checkers []servicequota.Checker
+			checkers = append(checkers,
+				servicequota.NewVpcCheck(o.Err, o.CloudProvider, config.DefaultRequiredVpcs),
+				servicequota.NewEipCheck(o.Err, o.CloudProvider, config.DefaultRequiredEpis),
+				servicequota.NewIgwCheck(o.Err, o.CloudProvider, config.DefaultRequiredIgws))
+			servicequota.CheckQuotas(checkers)
```
paulbes, line 260: You could do this inline if `CheckQuotas` has a variadic receiver, e.g, `CheckQuotas(checkers ...Checker)`: https://gobyexample.com/variadic-functions

paulbes (commented):
Looks great! Only one minor comment that you can choose to ignore :D

cmd/okctl/create.go
ivaruf: Looks better :) Done!

PR #86, opened 2020-09-23
[KM71] Change from Github Oauth App to AWS Cognito User Pool 

paulbes:
**NB:** This is still missing some pieces

Creating Github Oauth Apps requires a lot manual steps, AWS Cognito does not.

## Description
Github does not provide an API for creating an Oauth App, in addition, the app needs to be transferred to the main organisation, which is an additional manual step. To make the matters worse, we need to repeat this process for each Oauth app we want to create, e.g., grafana, argocd, etc.

With AWS Cognito we can create a user pool and as many clients we want without having to rely on user input. This makes the whole process a lot smoother.

The only setback is figuring out how to make it easy to sync teams between github and those in the cognito user pool.

## Types of changes

## Checklist:


pkg/api/core/cloudprovider/aws/identitymanager.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/oslokommune/okctl/pkg/api"
+	"github.com/oslokommune/okctl/pkg/api/okctl.io/v1alpha1"
+	"github.com/oslokommune/okctl/pkg/cfn"
+	"github.com/oslokommune/okctl/pkg/cfn/components"
+	"github.com/oslokommune/okctl/pkg/cognito"
+)
+
+const (
+	// DefaultCloudFrontHostedZoneID is the default hosted zone id for all cloud front distributions
+	DefaultCloudFrontHostedZoneID = "Z2FDTNDATAQYW2"
```
ivaruf, line 17: Will this be a problem for multiple hosted zones on same AWS account?

paulbes: See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-aliastarget.html

> CloudFront distribution
Specify Z2FDTNDATAQYW2. This is always the hosted zone ID when you create an alias record that routes traffic to a CloudFront distribution.

PR #93, opened 2020-09-28
Km76 Scaffold ArgoCD deployment from an application.yaml

deifyed:
## Description

`okctl create application prod` will create an application.yaml template. After configuring the application.yaml to suit your application, you can use it to scaffold the required resources needed to deploy by running
`okctl apply application prod -f application.yaml`

## Motivation and Context

To simplify the process of deploying an app. For a beginner, creating Kubernetes resources can be quite overwhelming. For experienced Kubernetes user there is time to be saved. For anyone not familiar with ArgoCD its nice to get some help along the way.

image
image
image
image

## Types of changes

## Checklist:

## TODO

- [x] Figure out if Application.Source.RepoURL is the IAC repo or the application/service's repo
- [x] Add required EKS Ingress annotations
- [x] Configure the Ingress URL correctly


cmd/okctl/applyapplication.go
```diff
@@ -0,0 +1,124 @@
...
+	return err
+}
+
+func handleOutput(o *okctl.Okctl, app api.Application, outputFormat string, expandedApp *okctlapplication.ArgoCDDeploymentResources) error {
+	switch outputFormat {
+	case "stdout":
+		fmt.Fprint(o.Out, &expandedApp.KubernetesResourcesBuffer)
+		fmt.Fprint(o.Out, &expandedApp.ArgoAppBuffer)
+	case "files":
+		rootPath := filepath.Join("./deployment", app.Name)
```
paulbes, line 99: We should probably place this under the `output` directory that the user has defined, e.g., `o.GetRepoOuputDir()` (can't remember the specific name, but something to this effect)

pkg/okctlapplication/utils.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+
+	"github.com/oslokommune/okctl/pkg/config"
+	"github.com/oslokommune/okctl/pkg/config/load"
+	"github.com/oslokommune/okctl/pkg/config/state"
+	"github.com/oslokommune/okctl/pkg/okctl"
+	"github.com/spf13/cobra"
+)
+
+func loadRepoData(o *okctl.Okctl, cmd *cobra.Command) error {
```
paulbes, line 13: What problem made you replicate this function? It is not entirely unlikely that we are loading user and repository data in a sub-optimal way. If that is the case, we should probably do something about it.

pkg/okctlapplication/utils.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/*
+GetCluster returns a certain cluster if env is specified and DefaultRepositoryConfig is available.
+If no env is specified and DefaultRepositoryConfig is available, GetCluster will return the first
+cluster it finds.
+
+If theres no DefaultRepositoryConfig available, GetCluster returns nil
+*/
+func GetCluster(o *okctl.Okctl, cmd *cobra.Command, env string) *state.Cluster {
```
paulbes, line 32: We have the `RepoStateFromEnv` that contains this data: https://github.com/oslokommune/okctl/blob/master/pkg/config/config.go#L165, you can pass this in to your struct, using the specific interface you need, e.g., `Clusterer` (https://github.com/oslokommune/okctl/blob/master/pkg/config/state/repositorysaver.go#L20)

pkg/okctlapplication/kaexintegration.go
```diff
@@ -0,0 +1,83 @@
...
+AcquireApplication returns an okctl Application based on stdin or a file
+*/
+func AcquireApplication(path string) (api.Application, error) {
+	var (
+		rawApplication []byte
+		err            error
+	)
+
+	if path == "-" {
+		rawApplication, err = ioutil.ReadAll(os.Stdin)
```
paulbes, line 23: We should pass in okctl's `In` here, instead of `os.Stdin`, https://github.com/oslokommune/okctl/blob/master/pkg/context/context.go#L28. This makes it easier to test. We could also use the `io.Reader` interface for both of these branches in the if statement, e.g.:

```go
r = os.Open(...) // returns an io.Reader
```

Then we could read in the bytes from `r`, which can be either an opened file or `os.Stdin`, or whatever else implements the `io.Reader` interface.

cmd/okctl/createapplication.go
```diff
@@ -0,0 +1,93 @@
...
+	opts := &CreateApplicationOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "application",
+		Short: "Create an application template",
+		Long:  "Scaffolds an application.yaml template which can be used to produce necessary Kubernetes and ArgoCD resources",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			var buffer bytes.Buffer
+
+			if opts.fullExample {
```
paulbes, line 43: This should be passed in as an argument to `scaffold.NewApplication(...)` or something to this effect and then handled by the internals of that implementation

paulbes (commented):
Good stuff, there are a lot of `okctl` internal packages that could be leveraged, I have tried to add comments where I found likely candidates. Not surprising that you haven't found them all yet :D

One possible area of improvement, I think the code might become cleaner if a constructor of a struct with methods was introduced, e.g.:

```go
package scaffold

type Persister interface { // We can implement this with a filesystem persister, in-memory persisted or io.Writer persister, or something to this effect
  Write([]byte) error
}

// Not sure about these functions, but I hope the idea comes across?
type ApplicationCreator interface { // bad name, but something to this effect
  Fetch(...) ...
  Persist(...) ...
}

type Application struct {
  In io.Reader // for okctl.In
  Persister Persister // an interface that defines how we will persist stuff
  Full bool // in Fetch we can then determine what to return
  Cluster state.Clusterer // for state interaction
}

func NewApplication(...) *Application {
  return &Application{...}
}
```

cmd/okctl/applyapplication.go
```diff
@@ -0,0 +1,124 @@
...
+
+		err := ensurePathExistence(rootPath)
+		if err != nil {
+			return fmt.Errorf("unable to create required folder structure: %w", err)
+		}
+
+		kubernetesResourcesPath := path.Join(rootPath, fmt.Sprintf("%s.yaml", app.Name))
+		argoAppResourcePath := path.Join(rootPath, fmt.Sprintf("%s-application.yaml", app.Name))
+
+		err = ioutil.WriteFile(kubernetesResourcesPath, expandedApp.KubernetesResourcesBuffer.Bytes(), 0o600)
```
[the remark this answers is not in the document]
deifyed: @paulbes Why should it be?

pkg/okctlapplication/utils.go
deifyed: I prefer to not fail unless it is impossible not to. I replicated the function so that in the scenario where there is no .okctl available, I'll just populate with dummy data. The alternative (as far as I know) will start the wizard, which will abort the command if a user chooses to cancel the wizard

paulbes: Is this because asking for user input messes up the piping of data?

cmd/okctl/applyapplication.go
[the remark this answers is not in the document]
paulbes: Just curious, since on most nix systems the default file permission is `0666` with a umask of `0022`, which gives `0644`. Personally, if there is no sensitive content within the file I see no reason to deviate from this behavior? I guess for background processes running as non-root/user it could negate some possibilities?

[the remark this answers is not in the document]
deifyed: Good enough for me. I just used what an example said which looked safe. 0644 it is

pkg/okctlapplication/utils.go
deifyed: Mostly because I by default don't want to force the user into some kind of pattern. Might be an erronous assumption in this case considering our opinionated approach, I just find it a neat rule to live by.

deifyed:
Closing this in favor of a more cleaned up PR series

PR #94, opened 2020-09-29
[KM79] Tell user about git repo requirement

yngvark:
Having an existing git repo is a prerequisite.

## Description
There is currently little information to the user about the prerequisite that `okctl create cluster` assumes it's running in a git repository. The only indication for the user is that when running `ok ctl create cluster test 123` you get an error message that includes a raw error message from git about a missing repository, but no explanation as to why the repository is missing.

Furthermore, the okctl.io's "Basic usage" doesn't mention that you need to run the `okctl create cluster` command from a git repository, so everyone is going to get an error message when attempting to use okctl create cluster for the first time. Since all the user is doing is following the instructions from okctl.io to the point, this shouldn't lead to an error.

Solution:
* Put all required commands to okctl.io's "Basic usage", so they work and don't produce an error by default
* Rewrite the error message when running `okctl create cluster` outside a git repository to be more human friendly
* Add short explanation in the description of the `okctl create cluster` about the git repository prerequisite

Let me know if the suggested changes in https://github.com/oslokommune/okctl/pull/94/files#diff-1b83d055aaa0d7c619f10fbb64014e85R38-R39 makes the instructions look horrible.

## How Has This Been Tested?

I tested manually by actually having to temporarily move the `okctl/.git` away, so that I could run okctl without being in a git directory.

Before this PR, the following instructions will lead to an error:

(From okctl.io's front page)
image

image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/create.go
+++ b/cmd/okctl/create.go
@@ -45,7 +45,7 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create commands",
-		Long: `Create various kinds of clusters.
+		Long: `Create various kinds of resources.
 
 Pre-requisites:
 * okctl create infrastructure-as-code files, and assumes that it is
--- a/cmd/okctl/create.go
+++ b/cmd/okctl/create.go
@@ -45,6 +45,12 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create commands",
+		Long: `Create various kinds of clusters.
+
+Pre-requisites:
+* okctl create infrastructure-as-code files, and assumes that it is
+running in a git repository. Initialize or clone a git repository
+before running any of these commands.`,
 	}
 
 	cmd.AddCommand(buildCreateClusterCommand(o))
--- a/cmd/okctl/create.go
+++ b/cmd/okctl/create.go
@@ -50,7 +50,7 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
 		Long: `Create various kinds of resources.
 
 Pre-requisites:
-* okctl creates various configuration files, and assumes that it is
+okctl creates various configuration files, and assumes that it is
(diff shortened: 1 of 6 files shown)


cmd/okctl/create.go
```diff
@@ -45,6 +45,12 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create commands",
+		Long: `Create various kinds of clusters.
```
deifyed, line 48: I'm currently adding a `okctl create application` command, which is not a cluster and will make this description misleading

cmd/okctl/create.go
```diff
@@ -45,6 +45,12 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
...
+
+Pre-requisites:
+* okctl create infrastructure-as-code files, and assumes that it is
```
deifyed, line 51: Ref my message above + a typo s/create/creates/

cmd/okctl/create.go
yngvark: Let's see who merges first :trollface: 

OK, changed to `resources`.

cmd/okctl/create.go
yngvark: Changed to `okctl creates various configuration files`.

cmd/okctl/create.go
```diff
@@ -47,6 +47,12 @@ func buildCreateCommand(o *okctl.Okctl) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create commands",
+		Long: `Create various kinds of resources.
+
+Pre-requisites:
+* okctl creates various configuration files, and assumes that it is
```
ivaruf, line 53: Remove the '*' in the start of this sentence (?)

ivaruf (approved):
Looks good :+1:

cmd/okctl/create.go
yngvark: I was thinking about this point as a list. Although there's just element now, so I can remove the `*` and possibly re-add it if we ever get one more requirement.

PR #106, opened 2020-10-20
KM71-1 Add add user pool user

ivaruf:
Adduser command 

NOTE: This PR is split into three parts 1 / 2 and 3 (final), each branch is based on the previous, so only the final branch needs to merge once all has been approved.

## Description
Gives an okctl user ability to add an idedentitypool user with the new adduser command

## Motivation and Context
After we changed from Oauth, okctl users need to add their own users into an identity-pool

## How Has This Been Tested?
Added new tests, and tested manually

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] I have updated the release notes (for the next release).


yngvark (approved):
Everything looks ok. However, if there is time for it, could we do a walkthrough in a meeting? As I'm quite green yet, I get what each function sort of does, but it's hard to get a feel for the whole picture.

ivaruf:
> Everything looks ok. However, if there is time for it, could we do a walkthrough in a meeting? As I'm quite green yet, I get what each function sort of does, but it's hard to get a feel for the whole picture.

Yes, we can do this for sure. Was thinking this also, or alternatively we make new functionality that also go through the entire client-api stack with @paulbes

PR #108, opened 2020-10-20
Km71-3 (final) Adduser fix tests and add release notes

ivaruf:
See comments in #106


userdocs/src/usage/adduser.md
```diff
@@ -0,0 +1,8 @@
+
+Add users into identitypool that will be used to loging to ArgoCD and other services
```
yngvark, line 2: Like in my comment in https://github.com/oslokommune/okctl/pull/107, the word identitypool might not be anything the user is familiar with. Can we explain it some other way? Or explain identitypool?

yngvark (changes requested):

userdocs/src/usage/adduser.md
ivaruf: Hmm, this is not user-documentation. I think it is fine.

yngvark: It is in the folder called "userdocs", and is on our web page. But we could suffice with just writing "AWS identitypool" so people get something to google.

yngvark (approved):

yngvark (approved):

PR #109, opened 2020-10-20
Add get team members from github

ivaruf:
## Description
Get team members using github api

## Motivation and Context
We need to be able to sync identitypool-users with github teams, this is a standalone step 
that just fetches the users from giyhub

## How Has This Been Tested?
Tested manually

## Types of changes
- [X] Non breaking new functionality that will be used later

## Checklist:

### DIFF
--- a/pkg/client/core/api/rest/github.go
+++ b/pkg/client/core/api/rest/github.go
@@ -20,6 +20,15 @@ type githubAPI struct {
 	out          io.Writer
 }
 
+func (a *githubAPI) GetGithubTeamMembers(opts client.GetGithubTeamMembers) (*[]client.GithubTeamMember, error) {
+	members, err := a.client.ListTeamMembers(*opts.Team)
+	if err != nil {
+		return nil, err
+	}
+
+	return &members, nil
+}
+
 func (a *githubAPI) SelectGithubInfrastructureRepository(opts client.SelectGithubInfrastructureRepositoryOpts) (*client.SelectedGithubRepository, error) {
 	repos, err := a.client.Repositories(opts.Organisation)
 	if err != nil {
@@ -90,6 +99,7 @@ func (a *githubAPI) SelectGithubTeam(opts client.SelectGithubTeam) (*client.Gith
 		ID:           opts.ID,
 		Organisation: opts.Organisation,
 		Name:         team.GetName(),
+		TeamID:       *team.ID,
 	}, nil
 }
 
--- a/pkg/client/github.go
+++ b/pkg/client/github.go
@@ -130,11 +130,27 @@ type GithubTeam struct {
 	ID           api.ID
 	Organisation string
 	Name         string
+	TeamID       int64
 }
 
 // Validate the data
 func (t GithubTeam) Validate() error {
 	return validation.ValidateStruct(&t,
+		validation.Field(&t.Name, validation.Required),
+	)
+}
+
+// GithubTeamMember info about
+type GithubTeamMember struct {
+	Name  string
+	Email string
+	Login string
+}
+
+// Validate the github team memmber
+func (t GithubTeamMember) Validate() error {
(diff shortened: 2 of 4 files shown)


pkg/client/github.go
```diff
@@ -130,6 +130,7 @@ type GithubTeam struct {
 	ID           api.ID
 	Organisation string
 	Name         string
+	TeamID       int64
```
yngvark, line 133: Just double checking: Is team ID actually an int and not a string? If team ID starts with zero, and is converted to an int, and then back to a string, the zero will be lost.

pkg/github/github.go
```diff
@@ -18,6 +20,7 @@ type Githuber interface {
 	Teams(org string) ([]*Team, error)
 	Repositories(org string) ([]*Repository, error)
 	CreateDeployKey(org, repository, title, publicKey string) (*Key, error)
+	ListTeamMembers(team client.GithubTeam) ([]client.GithubTeamMember, error)
```
yngvark, line 23: Is there any reason it's called ListTeamMembers here, vs GetGithubTeamMembers above (list vs get)?

yngvark (changes requested):
So this PR just adds the get functionality, another PR will actually do the sync, right?

pkg/client/github.go
ivaruf: Yepp, its actually an int :) 
See
go-github/v32 ...teams.go

```
// Team represents a team within a GitHub organization. Teams are used to
// manage access to an organization's repositories.
type Team struct {
	ID          *int64  `json:"id,omitempty"`
```

ivaruf: And yepp, the sync will come later . Waited with that since most people in Origo do not have a public facing e-mail on their git-hub account (yet).

pkg/github/github.go
ivaruf: Hmm, well - it takes a team as an argument, and does not really "get" anything, just some putting things in correct objetcs and returning.
Also, underlying command in github package is called ListTeamMembersByID

yngvark (approved):

pkg/client/core/api/rest/github.go
```diff
@@ -20,6 +20,15 @@ type githubAPI struct {
 	out          io.Writer
 }
 
+func (a *githubAPI) GetGithubTeamMembers(opts client.GetGithubTeamMembers) (*[]client.GithubTeamMember, error) {
```
paulbes, line 23: You probably want to return a slice of pointers to a GithubTeamMember, and not a pointer to a slice?

pkg/github/github.go
```diff
@@ -26,6 +29,42 @@ type Github struct {
...
+	teamID := team.TeamID
+
+	org, _, err := g.Client.Organizations.Get(ctx, team.Organisation)
+	if err != nil {
+		return nil, err
+	}
+
+	users, _, err := g.Client.Teams.ListTeamMembersByID(ctx, *org.ID, teamID, nil)
+	if err != nil {
+		fmt.Println(err)
```
paulbes, line 44: I know this is WIP, but just adding a note here, should return the err here

pkg/github/github.go
```diff
@@ -26,6 +29,42 @@ type Github struct {
...
+	}
+
+	members := []client.GithubTeamMember{}
+
+	for _, u := range users {
+		// Yes we really do need to do a call for every user,
+		user, _, _ := g.Client.Users.Get(ctx, *u.Login)
+
+		email := ""
+		if user.Email != nil {
```
paulbes, line 54: I think the github client has a getter for each of the fields, so you can use:`GetEmail`, which will do this for you. At least I have some recollection of this

paulbes (commented):
WIP I know, but some feedback

ivaruf:
This will not be needed it seems.

PR #114, opened 2020-10-23
KM76 feat(kaex) Part 1: Add kaex integration API

deifyed:
## Description

This PR contains the actual Kaex integration API package. The following PR's builds on this.

## Motivation and Context

This establishes the base which upon we'll build the scaffolding commands on.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (In a future PR)

### DIFF
--- a/pkg/scaffold/ingress.go
+++ b/pkg/scaffold/ingress.go
@@ -1,0 +1,55 @@
+package scaffold
+
+import (
+	"fmt"
+	"net/url"
+
+	kaex "github.com/oslokommune/kaex/pkg/api"
+	networkingv1 "k8s.io/api/networking/v1beta1"
+	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// CertificateCreatorFn handles generating a certificate for a certain host
+type CertificateCreatorFn func(fqdn string) (string, error)
+
+func createOkctlIngress(app kaex.Application, certificateCreatorFn CertificateCreatorFn) (*networkingv1.Ingress, error) {
+	ingress, err := kaex.CreateIngress(app)
+	if err != nil {
+		return nil, err
+	}
+
+	ingress.Spec.Rules[0].HTTP.Paths[0].Path = "/*"
+
+	if ingress.Annotations == nil {
+		ingress.Annotations = map[string]string{}
+	}
+
+	ingress.Annotations["kubernetes.io/ingress.class"] = "alb"
+	ingress.Annotations["alb.ingress.kubernetes.io/scheme"] = "internet-facing"
+
+	if ingress.Spec.TLS != nil {
+		parsedURL, _ := url.Parse(app.Url)
+
+		certificateARN, err := certificateCreatorFn(parsedURL.Host)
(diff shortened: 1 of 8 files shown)


pkg/scaffold/ingress.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"net/url"
+
+	kaex "github.com/oslokommune/kaex/pkg/api"
+	networkingv1 "k8s.io/api/networking/v1beta1"
+	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// CertificateCreatorFn handles generating a certificate for a certain host
```
yngvark, line 12: Can we describe what the parameters and return values are?

pkg/scaffold/template.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return buffer.Bytes(), nil
+}
+
+// InterpolationOpts defines possible data to inject into the templates
+type InterpolationOpts struct {
+	Domain string
+}
+
+// InterpolateTemplate replaces dummy data in the template with state dependant data
```
yngvark, line 36: I'm a bit unsure by reading this doc if this function modifies the passed byte array or not, and what is the return value.

pkg/scaffold/template.go
```diff
@@ -0,0 +1,76 @@
...
+func InterpolateTemplate(template []byte, opts *InterpolationOpts) ([]byte, error) {
```
yngvark, line 37: In engineering and science, the word [interpolate](https://en.wikipedia.org/wiki/Interpolation) is a method of estimation. For instance in the number series "1 2 x 4 5", we can interpolate that x=3, but it could very well actually be something completely different. In this function we are not estimating anything, we are replacing values.

However, I see the [dictionary](https://www.merriam-webster.com/dictionary/interpolate) says "to insert between other things or parts", which doesn't sound too far off, except that it says insert and not replace.

if you follow these thoughts, we should consider some other naming. What do you think?

yngvark (changes requested):
Looks clean & nice

ivaruf (approved):

pkg/scaffold/template.go
deifyed: I see the term interpolation used when dealing with templates to inject values where there is variables, thats why I named it that.

Ref: https://medium.com/@channaly/string-interpolation-in-golang-ecb3bcb2d600
https://intellij-support.jetbrains.com/hc/en-us/community/posts/360007837900-How-do-you-toggle-template-interpolation-in-Helm-chart-templates-
https://angular.io/guide/interpolation
https://pugjs.org/language/interpolation.html
etccccc

I felt it hit the nail on the head, but sure - if its a confusing term I can change it. Want me to change it? @yngvark @ivaruf could chime in as well

yngvark: Okay, it seems string interpolation is an established Thing and definitely what you have used. Keep it :+1:

yngvark (approved):

PR #115, opened 2020-10-23
KM76 feat(kaex) Part 2: Add service layer 

deifyed:
## Description

Establishes a layer in between the Kaex integration and okctl to make application scaffolding pluggable

## Motivation and Context

Easier to interchange functionality in the future. Ensures the same pattern is followed for the Kaex integration as elsewhere

## Types of changes

## Checklist:


pkg/client/core/report/console/application.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/oslokommune/okctl/pkg/client/store"
+	"github.com/oslokommune/okctl/pkg/spinner"
+)
+
+type applicationReport struct {
+	console *Console
+}
+
+// ReportDeleteApplication produces relevant output upon application deletion
+func (r *applicationReport) ReportDeleteApplication(reports []*store.Report) error {
```
yngvark, line 17: Go question which I maybe should post somewhere else: Since ReportDeleteApplication is a method defined on the interface client.ApplicationReport, shouldn't the documentation be placed on the interface instead of this specific method? The comment for this function will be readonly only by consumers in the same package.

Perhaps both could be documented if there was a slight difference between an applicationReport and a client.ApplicationReport.

pkg/client/core/report/console/application.go
```diff
@@ -0,0 +1,43 @@
...
+
+	for _, report := range reports {
+		actions = append(actions, report.Actions...)
+	}
+
+	return r.console.Report(actions, "application", aurora.Green("removing").String())
+}
+
+// ReportCreateApplication produces relevant output upon application creation
+func (r *applicationReport) ReportCreateApplication(application *client.ScaffoldedApplication, reports []*store.Report) error {
```
yngvark, line 28: Same as above.

pkg/client/core/service_application.go
```diff
@@ -0,0 +1,154 @@
...
+func writeSuccessMessage(writer io.Writer, applicationName string, argoCDResourcePath string) {
+	fmt.Fprintf(writer, "Successfully scaffolded %s\n", applicationName)
+	fmt.Fprintln(writer, "To deploy your application:")
+	fmt.Fprintln(writer, "\t1. Commit and push the changes done by okctl")
+	fmt.Fprintf(writer, "\t2. Run kubectl apply -f %s\n", argoCDResourcePath)
+	fmt.Fprintf(writer, "If using an ingress, it can take up to five minutes for the routing to configure")
+}
+
+// ScaffoldApplication turns a file path into Kubernetes resources
+func (s *applicationService) ScaffoldApplication(ctx context.Context, opts *client.ScaffoldApplicationOpts) error {
```
yngvark, line 56: Let's say this crashes in one of the steps, and the user attempts to scaffold the application again - is this function idempotent (and thus continuable, doesn't create a new set of certificats or other resources that needs to be cleaned up)?

yngvark (changes requested):

pkg/client/core/report/console/application.go
ivaruf: Exported functions are documented in the implementation, it seems this is how the linter also expects things to be, and I assume its part of how go generats docs ...

yngvark: If this is intended, it's fine. However, it seems strange to me to document the implementation (which only the package developers would read) and not the exported function, as people using your package are the ones I would imagine actually need documentation.

For instance, when using calling this method in [service_application.go](https://github.com/oslokommune/okctl/pull/116/files#diff-377a344d9c4a37a63ab3667b24a73fe7db6e97f12c8d100b76eeb0dd2aab1e56R36), you don't get any documentation.

deifyed: Makes sense @yngvark. I didn't really think about it. I'll move the comments

pkg/client/core/service_application.go
deifyed: Yeah pretty much. I know certificates are only created once, as long as the URL doesn't change.

I guess if writing the argocd resource should fail, the k8s resources will still have been written. I'm not about to implement transactions if thats what you're getting at :P (unless there is a function for that somewhere)

yngvark: No we don't want the complexity of that (now and here at least)! As long as it's okay to run this function 10 times, it's good.

yngvark (approved):

PR #124, opened 2020-10-30
Km72 delete cluster WIP 

ivaruf:
## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/delete.go
+++ b/cmd/okctl/delete.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/oslokommune/okctl/pkg/api/core/cleanup"
 	"regexp"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -122,6 +123,11 @@ including VPC, this is a highly destructive operation.`,
 
 			formatErr := o.ErrorFormatter(fmt.Sprintf("delete cluster %s", opts.Environment), userDir)
 
+			vpc, err := services.Vpc.GetVPC(o.Ctx, id)
+			if err != nil {
+				return formatErr(err)
+			}
+
 			// This is taken out, because of possible unintended consequences. The code is kept for now
 			/*
 				err = services.Domain.DeletePrimaryHostedZone(o.Ctx, o.CloudProvider, client.DeletePrimaryHostedZoneOpts{
@@ -137,6 +143,11 @@ including VPC, this is a highly destructive operation.`,
 				return formatErr(err)
 			}
 
+			err = services.ALBIngressController.DeleteALBIngressController(o.Ctx, id)
+			if err != nil {
+				return formatErr(err)
+			}
+
 			err = services.ExternalSecrets.DeleteExternalSecrets(o.Ctx, id)
 			if err != nil {
 				return formatErr(err)
@@ -147,14 +158,19 @@ including VPC, this is a highly destructive operation.`,
 				return formatErr(err)
 			}
 
-			err = services.ALBIngressController.DeleteALBIngressController(o.Ctx, id)
+			err = services.Cluster.DeleteCluster(o.Ctx, api.ClusterDeleteOpts{
+				ID: id,
+			})
 			if err != nil {
 				return formatErr(err)
 			}
 
-			err = services.Cluster.DeleteCluster(o.Ctx, api.ClusterDeleteOpts{
-				ID: id,
-			})
+			err = cleanup.DeleteDanglingALBs(o.CloudProvider, vpc.VpcID)
+			if err != nil {
+				return formatErr(err)
+			}
+
+			err = cleanup.DeleteDanglingSecurityGroups(o.CloudProvider, vpc.VpcID)
 			if err != nil {
 				return formatErr(err)
 			}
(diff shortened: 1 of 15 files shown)


cmd/okctl/delete.go
```diff
@@ -118,13 +117,15 @@ including VPC, this is a highly destructive operation.`,
 
 			formatErr := o.ErrorFormatter(fmt.Sprintf("delete cluster %s", opts.Environment), userDir)
 
-			err = services.Domain.DeletePrimaryHostedZone(o.Ctx, client.DeletePrimaryHostedZoneOpts{
+			err = services.Domain.DeletePrimaryHostedZone(o.CloudProvider, o.Ctx, client.DeletePrimaryHostedZoneOpts{
```
paulbes, line 120: FYI, this quote is from: https://golang.org/pkg/context/
> The Context should be the first parameter, typically named ctx

cmd/okctl/deletetestcluster.go
```diff
@@ -165,13 +165,20 @@ $ kubectl get service --all-namespaces
 				return formatErr(err)
 			}
 
+			// TODO --- yeah, this is hardcoded, can we get it from store?
 			err = services.Cluster.DeleteCluster(o.Ctx, api.ClusterDeleteOpts{
 				ID: id,
+				FargateProfileName: "fp-default",
```
paulbes, line 171: Unless something has changed, `eksctl` takes care of deleting the fargate profile as is. If something has gone wrong, it is most likely because ALBs are still active

cmd/okctl/deletetestcluster.go
```diff
@@ -165,13 +165,20 @@ $ kubectl get service --all-namespaces
...
 			})
 			if err != nil {
 				return formatErr(err)
 			}
 
+			err = services.ALBIngressController.DeleteAdditionalALBs(o.CloudProvider, o.Ctx, id)
```
paulbes, line 177: Deleting additional ALBs doesn't strictly have anything to do with the ALB ingress controller.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/aws-sdk-go v1.34.10
+	github.com/aws/aws-sdk-go-v2 v0.18.0
```
paulbes, line 19: `v2` is an experimental version of aws-sdk-go, and we shouldn't depend on both versions, prefer v1

pkg/client/core/api/rest/domain.go
```diff
@@ -12,6 +14,10 @@ type domainAPI struct {
 	client *HTTPClient
 }
 
+func (a *domainAPI) DeleteHostedZoneRecords(provider v1alpha1.CloudProvider, hostedZoneID string) error {
```
paulbes, line 17: This is very scary! Very, very, very scary. We need to be extremely careful, because we allow our users to reuse a hosted zone. There might be records in the zone that we do not "control". I don't think we want to do this with resources that we do not own. We should introduce an `okctl` tag and delete records related to that tag in which case.

paulbes (commented):
I know this is a work in progress, so some of these comments might not be valid / correct, etc. I am just dumping my initial thoughts here, without assuming how much you have thought of these things.

One thing that strikes me is that we should be using AWS tags more actively, e.g., we very much want to make sure we only delete things that are related to okctl, and specifically that cluster when performing these types of operations.\

We need to assume that resources can be more shared than we think.

cmd/okctl/deletetestcluster.go
ivaruf: I have read the source code. However, it never successfully actually manages to delete the cluster. If fargate profiles are removed up front. Then it works. :man_shrugging:

cmd/okctl/deletetestcluster.go
ivaruf: This is true, it will be moved :+1:

pkg/client/core/api/rest/domain.go
ivaruf: Hmmm --- maybe we should ask the user to manually delete any non SOA / NS records before running delete cluster ...

yngvark:
I'm getting this (release 0.0.28)

```bash
~/yk/git/oslokommune/yk6 (master) (dev:loki) $ okctl delete cluster dev
? This will delete yk6-dev and all assosicated resources, are you sure you want to continue? Yes
✓   deleting
Error: 
# Enable debug output
$ OKCTL_DEBUG=true okctl delete cluster dev
# Inspect the logs
$ cat /home/yngvar/.okctl/logs/console.log
# Ask for help on slack: #kjøremiljø-support

failed to process task Remove(): failed to remove file: remove /home/yngvar/yk/git/oslokommune/yk6/infrastructure/dev/identitypool: directory not empty
~/yk/git/oslokommune/yk6 (master) (dev:loki) $ OKCTL_DEBUG=true okctl delete cluster dev
DEBU[0000] binary already exists: eksctl (0.25.0)       
DEBU[0000] binary already exists: aws-iam-authenticator (0.5.1) 
DEBU[0000] binary already exists: kubectl (1.16.8)      
? This will delete yk6-dev and all assosicated resources, are you sure you want to continue? Yes
✓   deleting
Error: 
# Enable debug output
$ OKCTL_DEBUG=true okctl delete cluster dev
# Inspect the logs
$ cat /home/yngvar/.okctl/logs/console.log
# Ask for help on slack: #kjøremiljø-support

InvalidParameterException: No such domain or user pool exists.
```

PR #131, opened 2020-11-03
Delete all records in hosted zone before deletion, delete userpool

ivaruf:
## Description
Work in progress on the delete cluster command, now hosted zone, cluster and cognito pools will be deleted correctly.
Work remaining : proper deletion of ALB ingress controller and VPC

## Motivation and Context
We need to have a working delete command

## How Has This Been Tested?
Manual testing

## Types of changes

## Checklist:
- [ x] My code follows the code style of this project.
- [x ] All new and existing tests passed.

### DIFF
--- a/cmd/okctl/delete.go
+++ b/cmd/okctl/delete.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"regexp"
 
-	"github.com/AlecAivazis/survey/v2"
-
 	"github.com/oslokommune/okctl/pkg/client"
 
 	"github.com/oslokommune/okctl/pkg/spinner"
@@ -125,9 +123,14 @@ including VPC, this is a highly destructive operation.`,
 
 			formatErr := o.ErrorFormatter(fmt.Sprintf("delete cluster %s", opts.Environment), userDir)
 
-			err = services.Domain.DeletePrimaryHostedZone(o.Ctx, client.DeletePrimaryHostedZoneOpts{
+			err = services.Domain.DeletePrimaryHostedZone(o.Ctx, o.CloudProvider, client.DeletePrimaryHostedZoneOpts{
 				ID: id,
 			})
+			if err != nil {
+				return formatErr(err)
+			}
+
+			err = services.IdentityManager.DeleteIdentityPool(o.Ctx, o.CloudProvider, id)
 			if err != nil {
 				return formatErr(err)
 			}
--- a/cmd/okctl/delete.go
+++ b/cmd/okctl/delete.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/oslokommune/okctl/pkg/client"
 
 	"github.com/oslokommune/okctl/pkg/spinner"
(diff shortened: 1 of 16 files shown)


deifyed (approved):
LGTM. 

I did not test it

yngvark (approved):
This all looks good to me, though I know very little of the AWS specific stuff as well as what-goes-where in Okctl. Meaning I can't contribute with much 😄

cmd/okctl/delete.go
```diff
@@ -125,13 +124,18 @@ including VPC, this is a highly destructive operation.`,
 
 			formatErr := o.ErrorFormatter(fmt.Sprintf("delete cluster %s", opts.Environment), userDir)
 
-			err = services.Domain.DeletePrimaryHostedZone(o.Ctx, client.DeletePrimaryHostedZoneOpts{
+			err = services.Domain.DeletePrimaryHostedZone(o.Ctx, o.CloudProvider, client.DeletePrimaryHostedZoneOpts{
```
paulbes, line 127: I should have caught this last time, but we should not be injecting the `CloudProvider` here, this is a detail of the concrete implementation of the service itself.

We should rather be injecting the `CloudProvider` when creating the `services.Domain`

cmd/okctl/delete.go
```diff
@@ -125,13 +124,18 @@ including VPC, this is a highly destructive operation.`,
...
 				ID: id,
 			})
 			if err != nil {
 				return formatErr(err)
 			}
 
+			err = services.IdentityManager.DeleteIdentityPool(o.Ctx, o.CloudProvider, id)
```
paulbes, line 134: See another comment, but `CloudProvider` should not be injected here, but rather into the corresponding layer in the service.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/aws-sdk-go v1.34.10
+	github.com/aws/aws-sdk-go-v2 v0.18.0
```
paulbes, line 19: We should not be using two competing versions of `aws-sdk-go`, unless there is some really good reason for this?

pkg/client/core/api/rest/domain.go
```diff
@@ -12,6 +15,10 @@ type domainAPI struct {
 	client *HTTPClient
 }
 
+func (a *domainAPI) DeleteHostedZoneRecords(provider v1alpha1.CloudProvider, hostedZoneID string) (*route.ChangeResourceRecordSetsOutput, error) {
```
paulbes, line 18: Strictly speaking we should implement this delete functionality on the server-side, and then invoke the corresponding REST-API from this side

pkg/route53/route53.go
```diff
@@ -57,3 +59,51 @@ func (r *Route53) PublicHostedZones() ([]*HostedZone, error) {
 
 	return zones, nil
 }
+
+// DeleteHostedZoneRecordSets removes all recods in a hosted zone except NS and SOA records, to enable zone deletion
+func (r *Route53) DeleteHostedZoneRecordSets(hostedZoneID string) (*route.ChangeResourceRecordSetsOutput, error) {
```
paulbes, line 64: This makes me really nervous, I honestly believe we should introduce `okctl` tags in the records that we manage and only delete those. We do not want to end up in a situation where we wipe out all the records in a users hostedzone by mistake

paulbes (commented):
Many of the operations that we are performing client-side, due to the use of `CloudProvider` should be moved server-side and then rather invoke from the client

PR #135, opened 2020-11-05
KM96: Add "No PS1" option

yngvark:
Set env var `OKCTL_NO_PS1=true` to make `okctl venv` not set PS1.

DRAFT: Waiting for #132 to merge before creating this PR.

## Description

This makes the default setting be to set a PS1 for the user. The reason for this is to make it as simple as possible for beginners, which may not know their way around a terminal and custom PS1's too well - while at the same time give a good user experience.

Power users will have no problem putting this environment variable in their ~/.bashrc or similar without any issue.

## How Has This Been Tested?

Manually tested.

## Types of changes

## Checklist:


yngvark:
Due to rewrite of #132, this will be replaced with other PR(s).

PR #144, opened 2020-11-12
1/3 feat(venv): Prepare for venv rewrite, add package commands

yngvark:
PR 1/3 of
* PR 1: https://github.com/oslokommune/okctl/pull/144: Removes old stuff, add package commands
* PR 2: https://github.com/oslokommune/okctl/pull/145: Add package virtualenv
* PR 3: https://github.com/oslokommune/okctl/pull/146: Add tests

When done, I'll merge branch `96-add_ps1_to_venv_feature_branch` into master.

## About

Set PS1 (i.e. a custom command prompt) for the sub shell run by `okctl venv`.

## Description
When running `okctl venv myenv`, a sub shell is run. This PR sets the PS1 environment variable to show

```
<current dir> (<okctl environment:k8s namespace>)
```

See screenshot below.

More tests and documentation will be added in later PRs if this PR is accepted.

## Motivation and Context

* User feedback from demo stated that it would be nice to see some change when running `okctl venv`.
* We want the user not to run commands accidentally in the wrong k8s cluster, if he uses an existing k8s cluster already.

kube-ps1 does the same thing as already. However, to make using okctl as simple as possible, we come with this functionally pre-shipped.

The user may also want to use kube-ps1 instead. For this case, I'll make a new PR that adds the flag `--no-ps1` and an environment variable or something similar to `okctl venv`.

Options
* The user may want to use a custom PS1. In this case, s(he) can set OKCTL_PS1=your-own-ps1.
* The user  may want to keep her existing PS1. This is possible by setting OKCTL_NO_PS1=true

## How Has This Been Tested?

I wasn't sure if this PR would be accepted the way it is (the zsh hacks may be to hackish too accept). If it is accepted, I will make a new PR with more tests. I just didn't want to waste time.

I have tested this manually.

**BUG found** by @deifyed , __git_ps1 is not found when running in zsh. I'll fix that.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (See top comment about more PRs)
- [x] I have added tests to cover my changes. (Somewhat, see comment under test)


pkg/commands/venv.go
```diff
@@ -1,16 +1,168 @@
...
+		ClusterName:            cluster.Name,
+		UserDataDir:            userDataDir,
+		Debug:                  o.Debug,
+		KubectlBinaryDir:       path.Dir(k.BinaryPath),
+		AwsIamAuthenticatorDir: path.Dir(a.BinaryPath),
+	}
+
+	err = opts.Validate()
+	if err != nil {
+		return OkctlEnvironment{}, errors.E(err, "failed to validate show credentials options")
```
deifyed, line 78: This function is used in both `show credentials` and `venv` isn't it? This error message might prove to be confusing if it triggers in `venv`

pkg/commands/venv.go
```diff
@@ -1,16 +1,168 @@
...
+
+	if osPathExists {
+		return fmt.Sprintf("%s:%s", okctlPath, osPath)
+	}
+
+	return okctlPath
+}
+
+// MergeEnvVars first converts the given slice to a map. The provided slice must contain strings on the form "KEY=VALUE.
+// It then megres this map with the other provided map.
```
deifyed, line 132: Typo: merges*

pkg/commands/venv.go
```diff
@@ -1,16 +1,168 @@
...
+	osEnvMap := toMap(osEnvs)
+	for key, val := range osEnvMap {
+		merged[key] = val
+	}
+
+	// Merge PATHs
+	venvPath, venvHasPath := venvMap["PATH"]
+	osPath, osHasPath := osEnvMap["PATH"]
+
+	if osHasPath && venvHasPath {
```
deifyed, line 150: I'm just wondering in what situation would the venv PATH and the os PATH differ? Wouldn't they either be equal or the venv PATH empty?

userdocs/src/usage/venv.md
```diff
@@ -14,4 +14,27 @@ However, a simpler solution is to just run
...
+* `mynamespace` will be replaced with the name of the Kubernetes namespace of your current kube context
+
+This command prompt can be turned off or configured, see below.
+
+## Configuration
+
+| Environment variable      | Default   | Description                                                            |
+| ------------------        | --------- | ---------------------------------------------------------------------- |
+| OKCTL_NO_PS1              | true      | If true, disables overriding the PS1. This will keep the existing PS1. |
+| OKCTL_PS1                 | *not set* | If set, `venv` will use this as the PS1 in the executed subshell.      |
```
deifyed, line 37: Awesome improvement to documentation. I also love the OKCTL_PS1 option

deifyed (approved):

pkg/commands/venv.go
yngvark: Fixed, added new error msg

pkg/commands/venv.go
yngvark: Fixed

pkg/commands/venv.go
yngvark: The venv PATH will always differ, because the venv path contains the path kubectl and aws-iam-authenticator. Also, unless OKCTL_NO_PS1 is set, the venv path will contain the path to the PS1 executable (venv_ps1).

cmd/okctl/venv.go
```diff
@@ -30,9 +36,9 @@ export OKCTL_SHELL=/bin/bash
 okctl venv myenv
 `
 
-// nolint: funlen
+// nolint: gocyclo, funlen, gocognit
```
ivaruf, line 39: Is there any way to not ignore so many lint hints? (Split up command or similar?)
(I am a bit late with reviews, so if you are already ready to merge, feel free to ignore this)

ivaruf (approved):

cmd/okctl/venv.go
yngvark: Yeah, good points. I did so because I've till now been much less strict about the code in this package, because of the more nested nature of the cobra commands. But, that is not a good excuse really, so I have now extracted some functions to remove all the lint ignores.

PR #145, opened 2020-11-12
2/3 feat(venv): Add virtual environment

yngvark:
PR 2/3 of
* PR 1: https://github.com/oslokommune/okctl/pull/144
* PR 2: https://github.com/oslokommune/okctl/pull/145 
* PR 3: https://github.com/oslokommune/okctl/pull/146

PR #146, opened 2020-11-12
3/3 feat(venv): Add tests

yngvark:
PR 3/3 of
* PR 1: https://github.com/oslokommune/okctl/pull/144
* PR 2: https://github.com/oslokommune/okctl/pull/145 
* PR 3: https://github.com/oslokommune/okctl/pull/146

### DIFF
--- a/cmd/okctl/venv.go
+++ b/cmd/okctl/venv.go
@@ -36,7 +36,6 @@ export OKCTL_SHELL=/bin/bash
 okctl venv myenv
 `
 
-// nolint: gocyclo, funlen, gocognit
 func buildVenvCommand(o *okctl.Okctl) *cobra.Command {
 	okctlEnvironment := commands.OkctlEnvironment{}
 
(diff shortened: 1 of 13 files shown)


deifyed (approved):

ivaruf (approved):
Looks ready to merge :)

One thing : I see extensive work has been put into supporting zsh, would a mac user fall under the "bash" category, or unknown?
Edit: (Since you did such a good job supporting one type, now it is of course expected that you did the same for every common shell, hehe)

yngvark:
> Looks ready to merge :)
> 
> One thing : I see extensive work has been put into supporting zsh, would a mac user fall under the "bash" category, or unknown?
> Edit: (Since you did such a good job supporting one type, now it is of course expected that you did the same for every common shell, hehe)

If the user runs mac or linux shouldn't matter. The shell is retrieved from the user's login shell, which is found in /etc/passwd.

And yes, if okctl venv gets wildly popular I guess we will need to support more shells :grimacing: .

PR #167, opened 2020-11-24
[KM106]: Freeze golangci linter at v1.32.2

deifyed:
## Description

A bump in github.com/golangci/golangci-lint broke our CI. This PR freezes golang-lint at the last working version.

Changed to downloading binary due to warning about fetching
golangci-lint via `go get` here:
https://golangci-lint.run/usage/install/#install-from-source

To ensure it uses the new binary, it might be necessary to remove the following files:
* `${GOPATH}/bin/golangci-lint`
* `${GOPATH}/pkg/mod/github.com/golangci`

## How Has This Been Tested?

Manually with `make lint` and with CI by pushing to branch.

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,8 @@ GORELEASER     := bin/goreleaser/v.0.147.2/$(OS)/goreleaser
 GOFUMPT        := $(GOBIN)/gofumpt
 
 $(GOLANGCILINT):
+	# To bump, simply change the version at the end to the desired version. The git sha here points to the newest commit
+	# of the install script verified by our team located here: https://github.com/golangci/golangci-lint/blob/master/install.sh
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/b90551cdf9c6214075f2a40d1b5595c6b41ffff0/install.sh | sh -s -- -b ${GOBIN} v1.32.2
 
 $(GOIMPORTS):
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ GORELEASER     := bin/goreleaser/v.0.147.2/$(OS)/goreleaser
 GOFUMPT        := $(GOBIN)/gofumpt
 
 $(GOLANGCILINT):
-	$(GO) get github.com/golangci/golangci-lint/cmd/golangci-lint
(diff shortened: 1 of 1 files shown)


deifyed:
Something happens in 1.33.0. Couldn't figure out exactly what, so heres a PR for a temporary fix. Creating an issue for upgrading to 1.33.0

Makefile
```diff
@@ -75,7 +75,7 @@ GORELEASER     := bin/goreleaser/v.0.147.2/$(OS)/goreleaser
 GOFUMPT        := $(GOBIN)/gofumpt
 
 $(GOLANGCILINT):
-	$(GO) get github.com/golangci/golangci-lint/cmd/golangci-lint
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/b90551cdf9c6214075f2a40d1b5595c6b41ffff0/install.sh | sh -s -- -b ${GOBIN} v1.32.2
```
yngvark, line 78: It's a bit unclear what b90551cdf9c6214075f2a40d1b5595c6b41ffff0 points to. PR title says 1.32.2, shouldb't this url be `https://raw.githubusercontent.com/golangci/golangci-lint/v1.32.2/install.sh` then?

For both cases: Let's imagine somebody wants to update this version or wants to find out where this url comes from, it  would be nice with a link to where you found this url.

yngvark (changes requested):

Makefile
deifyed: @yngvark the reason for that line of code is in the commit message. I changed out the branch/release name to the actual commit since we're actually downloading and executing code here. Code in branches can change, release tags can change, but a commit reference can not change. Am I wrong?

yngvark: Good point! Defintely let it stay.

However, could we still add a comment so it's easy to update this commit hash?

deifyed: I'll add a comment describing where I got it from.

However that install script hasn't changed in the last 7 months, so its completely isolated from releases. To bump golangci-lint all we have to do is to change the last word on the line to the correct version

yngvark (approved):

PR #176, opened 2020-11-27
fix(venv): Run mac cmd to get login shell

yngvark:
Run `dscl . -read ~/ UserShell` to get the user's login shell on macOS, instead of trying to get it from /etc/passwd, which only works on Linux.

Browsing this PR: See separate commits to get a slightly smaller diff.

## Description

`okctl venv` didn't work on macOS, as it wrongfully assumed the user's login shell was found in /etc/passwd, which
is the case for Linux, but not macOS.

Unfortunately, I couldn't find a way to make this PR smaller, as everything is somewhat intertwined.

## How Has This Been Tested?

Added more tests.

Tested manually on a mac.

## Types of changes

## Checklist:


pkg/virtualenv/shellgetter/macos_login_shell_cmd_getter.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (*macOsLoginShellCmdGetter) parseShellCmd(line string) string {
+	split := strings.Split(line, ":")
+	afterColon := split[len(split)-1]
+	return strings.TrimSpace(afterColon)
+}
+
+// MacOsUserShellCmdGetter gets the user's login shell on macOS
+type MacOsUserShellCmdGetter interface {
```
yngvark, line 38: This is a wrapper for the exec.Command thing on line 53. We need this to be able to unit test the whole thing. Adds lots of extra code, though :\

pkg/virtualenv/shellgetter/os.go
```diff
@@ -0,0 +1,15 @@
+package shellgetter
+
+// Os specifies which operating system is being referred
+type Os int
+
+const (
+	// OsLinux represents the Linux OS
+	OsLinux Os = iota
```
ivaruf, line 8: what is iota?

ivaruf (approved):

pkg/virtualenv/shellgetter/os.go
yngvark: Some Go magic. When doing 

```go
const (
  a = iota
  b
  c
)

a will get the value 0
and b will automatically be 1, and c 2.

PR #179, opened 2020-11-27
KM94 feat(declarativecluster) PART 3: Add controller pkg API and usage

deifyed:
## Description

This PR implements the actually API we will use in okctl outside of the pkg/controller (`synchronize.go`) and the actual usage. It also includes the scaffold command that produces a cluster declaration template.

### DIFF
--- a/pkg/controller/synchronize.go
+++ b/pkg/controller/synchronize.go
@@ -59,11 +59,11 @@ func Synchronize(opts *SynchronizeOpts) error {
 
 	currentStateTree := CreateCurrentStateGraph(currentStateGraphOpts)
 
-	diffGraph := *opts.DesiredTree
+	diffTree := *opts.DesiredTree
 
-	diffGraph.ApplyFunction(applyCurrentState, currentStateTree)
+	diffTree.ApplyFunction(applyCurrentState, currentStateTree)
 
-	return handleNode(opts.ReconsiliationManager, &diffGraph)
+	return handleNode(opts.ReconsiliationManager, &diffTree)
 }
 
 // handleNode knows how to run Reconsile() on every node of a ResourceNode tree
--- a/pkg/controller/synchronize.go
+++ b/pkg/controller/synchronize.go
@@ -1,0 +1,93 @@
+package controller
+
+import (
+	"fmt"
+	"github.com/oslokommune/okctl/pkg/controller/reconsiler"
+	"github.com/oslokommune/okctl/pkg/controller/resourcetree"
+	"github.com/spf13/afero"
+)
+
+type SynchronizeOpts struct {
+	DesiredTree *resourcetree.ResourceNode
+	
+	ReconsiliationManager *reconsiler.ReconsilerManager
+	
+	Fs *afero.Afero
+	OutputDir string
+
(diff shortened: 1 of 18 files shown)


ivaruf (approved):
Looks very nice :+1:

pkg/controller/synchronize.go
```diff
@@ -0,0 +1,93 @@
...
+	diffGraph := *opts.DesiredTree
+
+	diffGraph.ApplyFunction(applyCurrentState, currentStateTree)
+
+	return handleNode(opts.ReconsiliationManager, &diffGraph)
+}
+
+// handleNode knows how to run Reconsile() on every node of a ResourceNode tree
+func handleNode(reconsilerManager *reconsiler.ReconsilerManager, currentNode *resourcetree.ResourceNode) error {
+	_, err := reconsilerManager.Reconsile(currentNode)
```
yngvark, line 71: Are we ignoring the ReconsilationResult because this being a MVP? What are possible uses of it in the future?

pkg/controller/synchronize.go
```diff
@@ -0,0 +1,93 @@
...
+	currentStateGraphOpts, err := NewCreateCurrentStateGraphOpts(opts.Fs, opts.OutputDir, opts.GithubGetter)
+	if err != nil {
+	    return fmt.Errorf("unable to get existing services: %w", err)
+	}
+
+	currentStateTree := CreateCurrentStateGraph(currentStateGraphOpts)
+
+	diffGraph := *opts.DesiredTree
+
+	diffGraph.ApplyFunction(applyCurrentState, currentStateTree)
```
yngvark, line 64: I just had to try to see how this would look without passing a function:

```go
func applyStateOnTree(desired *resourcetree.ResourceNode, current *resourcetree.ResourceNode) {
	for _, desiredChild := range desired.Children {
		applyStateOnTree(desiredChild, current)
	}

	targetNode := current.GetNode(desired)
	if desired.State == targetNode.State {
		desired.State = resourcetree.ResourceNodeStateNoop
	}
}
```

(Btw this code could be buggy, haven't tested it.)


This code more compact (as you don't need the diffGraph.ApplyFunction method anymore). But you lose the current tree iteration ability - getting an equivalent node from the target tree. I'm just wondering if you have considered simpler, but less future-proof alternatives - no need to change anything.

yngvark (changes requested):

pkg/controller/synchronize.go
deifyed: Ugh. This is me future proofing. 

However, the reconciliation result will be necessary when implementing the nameserver dance. I plan on creating a node which ensures the nameservers is in place - and this node will have to have ReconciliationResult{requeue: true} until the node sees that the nameservers has been added to the oslo.systems domain

pkg/controller/synchronize.go
deifyed: Yeah. I actually rewrote a function like that for the more future proof one.

I know that at some point soon I must implement delete cluster, in which I'll create a applyDeletedNodesInformation function and use the same "iterator"-function with. I hope that makes sense

yngvark (approved):

PR #180, opened 2020-11-27
KM94 feat(declarativecluster) PART 4: Add no user input modifications 

deifyed:
## Description

This PR contains the necessary changes I had to do to existing code to disable user input. Mainly the tactic here was to duplicate the existing functionality and remove Ask references.

### DIFF
--- a/pkg/config/load/user.go
+++ b/pkg/config/load/user.go
@@ -64,8 +64,6 @@ func CreateOnUserDataNotFoundWithNoInput() DataNotFoundFn {
 
 		data := state.NewUser()
 
-		data.User.Username = "ooo306134" // TODO: from env var?
-
 		c.UserState = data
 
 		_, err = store.NewFileSystem(userDir, c.FileSystem).
--- a/pkg/config/load/user.go
+++ b/pkg/config/load/user.go
@@ -50,6 +50,37 @@ all the time, if you so choose.
 
 `
 
+func CreateOnUserDataNotFoundWithNoInput() DataNotFoundFn {
+	return func(c *config.Config) (err error) {
+		userDir, err := c.GetUserDataDir()
+		if err != nil {
+			return fmt.Errorf("getting user data dir: %w", err)
+		}
+
+		userDataPath, err := c.GetUserDataPath()
+		if err != nil {
+			return fmt.Errorf("getting user data path: %w", err)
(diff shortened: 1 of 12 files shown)


pkg/config/load/user.go
```diff
@@ -50,6 +50,37 @@ all the time, if you so choose.
...
+		}
+
+		userDataPath, err := c.GetUserDataPath()
+		if err != nil {
+			return fmt.Errorf("getting user data path: %w", err)
+		}
+
+		data := state.NewUser()
+
+		data.User.Username = "ooo306134" // TODO: from env var?
```
ivaruf, line 67: Hard coded user?

pkg/okctl/okctl.go
```diff
@@ -746,6 +746,12 @@ https://www.passwordstore.org/
 		_ = k.Store(keyring.KeyTypeUserPassword, password)
 	}
 
+	//machineAuth := aws.NewAuthMachine( // TODO: make AuthMachine work and add it to 772:aws.New
```
ivaruf, line 749: Commented code?

ivaruf (changes requested):
Just a few small issues, see comments.

pkg/okctl/okctl.go
deifyed: Hmm. Woops. This is code from the auth part which should not be a part of this PR. I'll remove

pkg/config/load/user.go
deifyed: :D good catch

pkg/credentials/aws/aws.go
```diff
@@ -209,6 +209,69 @@ func DefaultStsProvider(sess *session.Session) stsiface.STSAPI {
 // to missing or expired credentials
 type PopulateFn func(*AuthSAML) error
 
+//type AuthMachine struct {
```
yngvark, line 212: Perhaps put in another branch instead of letting these comments go to master?

yngvark (changes requested):

pkg/credentials/aws/aws.go
deifyed: Yeah, I've removed them from this branch

yngvark (approved):

PR #183, opened 2020-11-30
Delete identity-pool from store on delete cluster

ivaruf:
## Description
Remove identity-pool from store on cluster deletion

## Motivation and Context
Should not be in store when it is deleted.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/client/core/store/filesystem/identitymanager.go
+++ b/pkg/client/core/store/filesystem/identitymanager.go
@@ -57,7 +57,7 @@ type IdentityPoolUser struct {
 	StackName  string
 }
 
-// RemoveIdentityPool removes alias, cert, and indentitypool files from file store
+// RemoveIdentityPool removes alias, cert, and identitypool files from file store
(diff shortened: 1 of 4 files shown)


pkg/client/core/store/filesystem/identitymanager.go
```diff
@@ -57,6 +57,29 @@ type IdentityPoolUser struct {
 	StackName  string
 }
 
+// RemoveIdentityPool removes alias, cert, and indentitypool files from file store
```
yngvark, line 60: indentitypool -> identitypool

unless there is some cool new aws service that indents for you

/funny

yngvark (approved):

pkg/client/core/store/filesystem/identitymanager.go
ivaruf: Bah! :D

PR #187, opened 2020-12-01
Km100 add delete hosted zone flag

ivaruf:
## Description
Enable a user to delete hosted zone on delete cluster

## Motivation and Context
Automatically deleting hosted zones are scary, by providing a flag that is hard to misunderstand, we enable powerusers and testers to delete hosted zone on cluster delete.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

- [] I have updated the documentation accordingly.

### DIFF
--- a/pkg/client/core/service_identitymanager.go
+++ b/pkg/client/core/service_identitymanager.go
@@ -35,6 +35,10 @@ func (s *identityManagerService) DeleteIdentityPool(ctx context.Context, provide
 
 	pool := s.state.GetIdentityPool()
 
+	if pool.UserPoolID == "" {
+		return nil
+	}
+
 	c := cognito.New(provider)
 
 	err = c.DeleteAuthDomain(pool.UserPoolID, pool.AuthDomain)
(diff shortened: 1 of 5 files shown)


pkg/client/core/service_identitymanager.go
```diff
@@ -35,6 +35,10 @@ func (s *identityManagerService) DeleteIdentityPool(ctx context.Context, provide
 
 	pool := s.state.GetIdentityPool()
```
ivaruf, line 37: Not really related, but if you try to run delete again and the user-pool is missing, it will continue instead of crashing ...

pkg/client/core/service_identitymanager.go
```diff
@@ -35,6 +35,10 @@ func (s *identityManagerService) DeleteIdentityPool(ctx context.Context, provide
...
 
+	if pool.UserPoolID == "" {
```
ivaruf, line 38: Not really related, but if you try to run delete again and the user-pool is missing, it will continue instead of crashing ...

yngvark (approved):
Should we add some release notes as well? Not sure about your preferred way of adding release notes.

PR #191, opened 2020-12-04
Unify userdoc with the same bash code layout for example commands

haavardeide:
## Description
Removes all `$` instances to avoid copying this into terminal when following step-by-step. Ensure that we use the same way of presenting bash code throught the userdocumentation

Correct layout of lists when generating userdoc

## Motivation and Context
When copying with the supplied `copy to clipboard` or selecting a entire line of code the `$` is included, and then have to be deleted before executing the command

## How Has This Been Tested?
`mkdocs serve`

### DIFF
--- a/userdocs/src/components/cloud.md
+++ b/userdocs/src/components/cloud.md
@@ -3,7 +3,7 @@
 ## Cloud components
 
 Cloud providers offer a vast array of functionality for:
- 
+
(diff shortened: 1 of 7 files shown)


ivaruf (approved):

ivaruf:
Looks good, thanks :)

PR #196, opened 2020-12-07
Userdocumentation: add example on how to add secret to Deployment

haavardeide:
## Description
Added example on how to add environment variables in a Deployment based on secret applied to kubernetes

### DIFF
--- a/userdocs/src/components/kubernetes.md
+++ b/userdocs/src/components/kubernetes.md
@@ -2,7 +2,7 @@
 
 ## Kubernetes components
 
-Kubernetes is [highly configurable and extensible](https://kubernetes.io/docs/concepts/extend-kubernetes/). We make use of this fact with `okctl` and bind our Kubernetes cluster setup tighter together with AWS. We do this by deploying a variety of applications into the cluster to manage various aspects of an application setup for us. 
+Kubernetes is [highly configurable and extensible](https://kubernetes.io/docs/concepts/extend-kubernetes/). We make use of this fact with `okctl` and bind our Kubernetes cluster setup tighter together with AWS. We do this by deploying a variety of applications into the cluster to manage various aspects of an application setup for us.
 
 - [Kubernetes External Secrets](#kubernetes-external-secrets)
 - [AWS ALB Ingress Controller](#aws-alb-ingress-controller)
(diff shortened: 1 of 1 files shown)


userdocs/src/components/kubernetes.md
```diff
@@ -48,6 +48,31 @@ data:
   admin_password: ...
 ```
 
+Inspect the secrets you just applied with:
+```bash
+kubectl get secrets
+kubectl describe secret postgres-config
+```
+or if you applied the secret to a namespace:
```
yngvark, line 56: You always apply a secret (or any Kubernetes resource) to a namespace, it's not possible not to.

We should say `or if you applied the secret to a specific namespace`

userdocs/src/components/kubernetes.md
```diff
@@ -36,7 +36,7 @@ spec:
       name: admin_password
 ```
 
-When this definition is applied to the cluster with `kubectl apply -f {secret.yaml}` it will result in a Kubernets secret being created.
+When this definition is applied to the cluster with `kubectl apply -f {secret.yaml}`, or if you deploy your pods to a namespace: `kubectl apply -f {secret-yaml} --namespace {your-namespace}`, it will result in a Kubernets secret being created.
```
yngvark, line 39: You always apply a pod (or any Kubernetes resource) to a namespace, it's not possible not to.

The reason `--namespace` often is omitted in docs, is that if you do kubectl apply <something> (without --namespace), it is going to apply to the current namespace. So it's not necessary to add `--namespace` to every command. However, not all users may know this, so it doesn't hurt to put it here. If we have many kubectl commands, though, I would rather mention how the user can set the active namespace in the start of the article, so we can avoid the `--namespace` argument everywhere.

Just FYI, the active namespace can be set like this:

```bash
kubectl config current-context // output example: my-context
kubectl config set-context my-context --namespace=mystuff
kubectl config use-context my-context
```

But a simpler alternative is to install for instance https://github.com/ahmetb/kubectx

Anyway, I suggest we write something like this:

When this definition is applied to the cluster with `kubectl apply -f {secret.yaml}`, or if you apply the definition to a specific namespace: `kubectl apply -f {secret-yaml} --namespace {your-namespace}`, it will result in a Kubernets secret being created.

yngvark (changes requested):
Good addition.

Needs some adjustment on how namespaces work.

paulbes:
This PR has been open for a while now, should we make a push to get it merged or close it?

yngvark:
I can just do the changes myself and merge it.

yngvark:
Closing in favour of #289.

PR #208, opened 2020-12-14
KM72: Delete cert in us-east-1 when deleting identity pool

ivaruf:
Sends a custom provider for us-east-1, to delete auth cert.

This code breaks client / api pattern. I need to go back and rewrite
some things later. Delete cluster needs to get done so, so it takes
priority over correctness.

## Description
Uses cloudformation to delete certifictate in us-east-1 on identity-pool deletion (delete cluster)

## Motivation and Context
We need to be able to clean up the resources we make

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.27.md
+++ b/docs/release_notes/0.0.27.md
@@ -11,7 +11,9 @@ By using the following commands, one can now create a cluster using a declaratio
 `okctl apply cluster` applies a cluster declaration that defines what parts of the cluster should exist
 
 For now, only resources that don't require a certificate will be created. 
+
 ## Bugfixes
+Remove certificate in us-east-1 when deleting identlity-pool as part of delete cluster.
 
 ## Other
 
--- a/pkg/api/core/cloudprovider/aws/identitymanager.go
+++ b/pkg/api/core/cloudprovider/aws/identitymanager.go
@@ -62,7 +62,7 @@ func (s *identityManagerCloudProvider) CreateIdentityPoolClient(opts api.CreateI
 		return nil, fmt.Errorf("retrieving identity pool client outputs: %w", err)
 	}
 
-	secret, err := cognito.New(s.provider).UserPoolClientSecret(c.ClientID, opts.UserPoolID)
+	secret, err := cognito.New(s.provider, nil).UserPoolClientSecret(c.ClientID, opts.UserPoolID)
 	if err != nil {
 		return nil, fmt.Errorf("retrieving client secret: %w", err)
 	}
@@ -97,7 +97,7 @@ func (s *identityManagerCloudProvider) CreateIdentityPool(certificateARN string,
 		return nil, fmt.Errorf("creating identity pool cloud formation stack: %w", err)
 	}
 
-	d, err := cognito.New(s.provider).UserPoolDomainInfo(opts.AuthDomain)
+	d, err := cognito.New(s.provider, nil).UserPoolDomainInfo(opts.AuthDomain)
 	if err != nil {
 		return nil, fmt.Errorf("getting cloudfront auth domain info: %w", err)
 	}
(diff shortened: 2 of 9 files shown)


pkg/cognito/cognito.go
```diff
@@ -2,16 +2,19 @@ package cognito
...
 	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
 	"github.com/oslokommune/okctl/pkg/apis/okctl.io/v1alpha1"
 )
 
 // Cognito contains all required state for interacting
 // with the Cognito API
 type Cognito struct {
-	provider v1alpha1.CloudProvider
+	provider   v1alpha1.CloudProvider
+	usprovider v1alpha1.CloudProvider
```
deifyed, line 17: I find this abit confusing. Shouldn't it be enough to send a single provider but instance it with an US region? Or are they functionality-wise different? In that case maybe we should name them with their functionality, like `creationProvider` and the `deletionProvider`.

deifyed (changes requested):

pkg/api/core/cloudprovider/aws/identitymanager.go
```diff
@@ -62,7 +62,7 @@ func (s *identityManagerCloudProvider) CreateIdentityPoolClient(opts api.CreateI
 		return nil, fmt.Errorf("retrieving identity pool client outputs: %w", err)
 	}
 
-	secret, err := cognito.New(s.provider).UserPoolClientSecret(c.ClientID, opts.UserPoolID)
+	secret, err := cognito.New(s.provider, nil).UserPoolClientSecret(c.ClientID, opts.UserPoolID)
```
yngvark, line 65: Sending nil is somewhat dangerous, as it may cause future panics, and we put the more of the responsibility to avoid this on the developer.

One way to avoid this can be to simply instantiate two cognito instances, one with eu-west provider, and one with us-east. However, if we want to emphasize that we only want one operation (`DeleteAuthCert`) on the us-east cognito, we can make two different structs, `EuWestCognito` and a `UsEastCognito`. The `UsEastCognito` only needs to contain the `DeleteAuthCert` method.

And maybe `UsCognito` should be named more after the functionality it provides, as @deifyed suggests. `CertDeleterCognito` or something.

But maybe this is overkill, thoughts?

docs/release_notes/0.0.27.md
```diff
@@ -11,7 +11,9 @@ By using the following commands, one can now create a cluster using a declaratio
 `okctl apply cluster` applies a cluster declaration that defines what parts of the cluster should exist
```
yngvark, line 11: Unrelated to this code line:

> This code breaks client / api pattern. I need to go back and rewrite
some things later. Delete cluster needs to get done so, so it takes
priority over correctness

As long as we agree that we actually will follow the client/api pattern later, and that later doesn't mean too long down the road (not sure which "deadline" which is good), I agree with the get-it-out-now decision.

yngvark (changes requested):

pkg/cognito/cognito.go
ivaruf: Nope, it is not possible to switch the region in a provider as it is made now, without running init agian. This is why I opted to having two of them

pkg/api/core/cloudprovider/aws/identitymanager.go
ivaruf: Hmm, went this route to avoid too much complexity, but i see your point - I'll try to make a new version of cognito instead :)

yngvark: We can ignore this for now just to get it out - but then we have to make sure not to forget it when rewriting it the future. If we make a task on it with this as a checkbox/todo, it should be good enough.

yngvark (approved):
Looks good!

deifyed (approved):

PR #212, opened 2020-12-17
KM72-Remove base folders in seperate step, that will not fail the run

ivaruf:
## Description
If some directories were not empty the delete proscess would fail

## Motivation and Context
We want delete to go on, even if a folder we figured would be empty still has files.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/client/core/store/filesystem/externalsecrets.go
+++ b/pkg/client/core/store/filesystem/externalsecrets.go
@@ -30,11 +30,14 @@ func (s *externalSecretsStore) RemoveExternalSecrets(_ api.ID) (*store.Report, e
 		Remove(s.chart.OutputFile).
 		Remove(s.chart.ReleaseFile).
 		Remove(s.chart.ChartFile).
-		Remove("").
 		Do()
 	if err != nil {
 		return nil, err
 	}
+
+	_, _ = store.NewFileSystem(s.chart.BaseDir, s.fs).
+		Remove("").
+		Do()
 
 	return report, nil
 }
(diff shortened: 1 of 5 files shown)


pkg/client/core/store/filesystem/externalsecrets.go
```diff
@@ -30,12 +30,15 @@ func (s *externalSecretsStore) RemoveExternalSecrets(_ api.ID) (*store.Report, e
 		Remove(s.chart.OutputFile).
 		Remove(s.chart.ReleaseFile).
 		Remove(s.chart.ChartFile).
-		Remove("").
 		Do()
 	if err != nil {
 		return nil, err
 	}
 
+	_, _ = store.NewFileSystem(s.chart.BaseDir, s.fs).
```
yngvark, line 38: In this and the other similar commands, are we not ignoring errors, which may cause to panic?

yngvark (approved):
See question.

pkg/client/core/store/filesystem/externalsecrets.go
ivaruf: We ignore errors, so it will just continue if there is an error

ivaruf: What happened last time was that delete execution stopped since a directory was not empty - and we had told it to return that error - now we just ignore it. 
The linter won't let you just
"store.NewFileSystem(s.chart.BaseDir, s.fs).Remove("").Do()"

yngvark: But we can't go around and just ignore errors, can we? :smile: If there's an error creating the filesystem, we can't just use as if nothing happened?

ivaruf: I agree, this ought to be rewritten, we have taken on some technical debt to get delete cluster functional :)

PR #231, opened 2021-01-06
fix(hostedzone): Set NS record's TTL to 900

yngvark:
## Description

Set TTL of NS record to 15 minutes when creating a hosted zone.

## Motivation and Context

When users delete a cluster, and creates it again with the same domain name, okctl create fails. This is due to DNS resolvers having cached a now outdated NS record for the domain. The step immediately after creating a hosted zone, create cognito identity pool, fails.

And because users frequently creates and deletes clusters when testing okctl for the first x times, they will run into this quite often.

The current problem before this PR is something like this (not 100% sure here): After a delete cluster, and when trying to create it again, oslo.systems's NS records correctly points to the new NS record of mycluster.oslo.systems. However, the NS record of the old mycluster.oslo.systems had TTL two days, meaning DNS resolvers _still_ will return the old NS record. So when okctl create (or anyone) attempts to use mycluster.oslo.systems for anything, DNS resolvers will ask the old NS name servers to do lookups, making any request fail to mynewcluster.oslo.systems fail. For instance, getting the A-record of myapp.mycluster.oslo.systems will resolve to the IP of myapp in my previous cluster.

By setting the TTL to 900 seconds, the user can just wait 15 minutes, instead of two days, after deleting a cluster before creating it again.

Why 900 secs / 15 minutes? Because this observation: https://blog.apnic.net/2019/11/12/stop-using-ridiculously-low-dns-ttls/  

> The number of required queries drops from 47% to 34% by setting a minimum TTL of 5 minutes, to 25% with a 15-minute minimum, and to 13% with a 1-hour minimum.

Our users can increase to 1 or 2 hours later when they are done testing, 15 minutes seems like a good sweet spot to facilitate multiple runs of okctl and a inefficient and too low TTL.

## How Has This Been Tested?

Manually.

image

## Types of changes

## Checklist:


deifyed (approved):
Looks good! 

Maybe we could add a default value to the NSTTL somewhere? Not sure if its needed, I'd just prefer to not have to read deep into DNS RFC's to figure out a sane value when using the create hosted zone function in the future

pkg/route53/route53.go
```diff
@@ -107,3 +109,61 @@ func (r *Route53) DeleteHostedZoneRecordSets(hostedZoneID string) (*route.Change
...
+	for _, set := range sets {
+		if *set.Type == "NS" {
+			nsCount++
+
+			nsSet = set
+		}
+	}
+
+	if nsCount != 1 {
+		return errors.Errorf("expected 1 NS records in hosted zone, but found %d", nsCount)
```
ivaruf, line 139: Don't we have 4 NS records in a hosted zone? ... Or is it maybe 4 values in the one records, yes, never mind ;)

ivaruf (approved):
Looks good :+1:

pkg/route53/route53.go
yngvark: Yes, the last one :100:

PR #233, opened 2021-01-07
Suggestion: suffix all duplicate name files so you can see where they come from 

ivaruf:
I am going crazy:

We have 8(!) files called "identitymanager.go" and two files called "service_identitymanager.go"

This PR is a suggestion for a new naming convention that will be a little easier on the mental load for such cases where we have
a lot of duplicate named files. (Basically suffixing the directory we are in)

Yes, I am aware that you can see which file it is based on the directory structure, I still think this change will make it easier to navigate.


ivaruf:
Since it is only renaming of files, go does not care what they are called - and git is able to pick up the rename changes and in my current branch (that does change a  lot of the "identitymanager.go" files ) with no conflicts.

<3 git
```
First, rewinding head to replay your work on top of it...
Applying: KM116: Add delete indentity-pool to handler
Applying: WERY WIP
Using index info to reconstruct a base tree...
A	pkg/api/core/cloudprovider/aws/identitymanager.go
A	pkg/api/core/service_identitymanager.go
A	pkg/api/identitymanager.go
A	pkg/client/core/service_identitymanager.go
A	pkg/client/identitymanager.go
.git/rebase-apply/patch:140: trailing whitespace.
	
warning: 1 line adds whitespace errors.
Falling back to patching base and 3-way merge...
```

deifyed (commented):
I agree that the current structure causes alot of mental friction.

However, I'm afraid the suggested prefixing will be too complicated and therefore cause alot of variation in convention down the line. 

For example; `client/core/store/filesystem/domain_filesystem.go` needs to have both store and client in its name, since `client/core/state/filesystem/domain_filesystem.go` is a possibility and will therefore clash. The same is a possibility for the API side of things.

Maybe the solution here can be to tune the IDE to better show the folder structure?

deifyed (approved):

PR #240, opened 2021-01-08
Km116 delete ssm and acm cert use api

ivaruf:
## Description
Clean up some usage in client that should be handeled by the api 

## Motivation and Context
Cleanup techdebt

## How Has This Been Tested?
Manually

## Types of changes

- [X] Clean and refactor

## Checklist:

- [ X] My code follows the code style of this project.
- [ X] All new and existing tests passed.
- [ X] I have updated the release notes (for the next release).

### DIFF
--- a/pkg/client/core/api/rest/certificate_rest.go
+++ b/pkg/client/core/api/rest/certificate_rest.go
@@ -12,6 +12,12 @@ type certificateAPI struct {
 	client *HTTPClient
 }
 
+// nolint: godox
+// TODO - Consider making own endpoint to delete certificate in separate
+func (a *certificateAPI) DeleteCertificate(opts api.DeleteCertificateOpts) error {
+	panic("Not implemented, called directly inside api when running delete identitypool")
+}
+
 func (a *certificateAPI) CreateCertificate(opts api.CreateCertificateOpts) (*api.Certificate, error) {
 	into := &api.Certificate{}
 	return into, a.client.DoPost(TargetCertificate, &opts, into)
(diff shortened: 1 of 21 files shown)


pkg/client/core/api/rest/certificate_rest.go
```diff
@@ -12,6 +12,12 @@ type certificateAPI struct {
 	client *HTTPClient
 }
 
+// nolint: godox
+// TODO - Consider making own endpoint to delete certificate in separate
```
deifyed, line 16: Undone TODO. Was this supposed to be a part of this cleaning? Or is it a TODO for the future

pkg/cognito/authcertdeleter.go
```diff
@@ -22,6 +22,10 @@ func (c *AuthCertDeleter) DeleteAuthCert(domain string) error {
 		return err
 	}
 
+	if domain == "" {
```
deifyed, line 25: I think we're supposed to put things like this into a struct as early as possible and find problems like this with struct.Validate()

deifyed (approved):
Code looks alot better! Good job

pkg/client/core/api/rest/certificate_rest.go
ivaruf: No, this will be a new task - the calls are made to the api in the apropriate fashion like this task calls for. This TODO marks that the logic should be moved and orchestrated in the client. Trying to keep the PR in a managable state :)

pkg/cognito/authcertdeleter.go
ivaruf: Yes, that would be better - think i need a new round of cleanup later, want this in for now

PR #267, opened 2021-01-28
feat(kaex): Add tests

deifyed:
## Description

- Adds test to `okctl scaffold application` to ensure interpolation and file writer works
- Adds test to `okctl apply application` to verify service functionality

## Motivation and Context

I didn't write tests when implementing this feature earlier, so it was about time

## How Has This Been Tested?

With tests.

## Types of changes

## Checklist:


pkg/client/core/service_application_client_test.go
```diff
@@ -0,0 +1,132 @@
+package core_test
```
yngvark, line 1: This whole file looks like a nice example for others to follow when creating a service test :+1:

yngvark (changes requested):
Exquisite tests!

PR #273, opened 2021-02-03
✅ Add service user auth

deifyed:
## Description

Enables using a service user through environment variables for okctl operations

## Motivation and Context

[Nightly testing](https://trello.com/c/k8LyFJLX/94-allow-for-programmatic-creation-of-clusters) requires being able to authenticate programmatically

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.39.md
+++ b/docs/release_notes/0.0.39.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- Allow for service user authentication through environment variables
+- Allow for AWS service user authentication through environment variables
 
 ## Bugfixes
 34fe9be: Fix ASM permissions for external secrets GH#243
--- a/docs/release_notes/0.0.39.md
+++ b/docs/release_notes/0.0.39.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Allow for service user authentication through environment variables
+
 ## Bugfixes
 34fe9be: Fix ASM permissions for external secrets GH#243
 
--- a/pkg/credentials/aws/aws.go
+++ b/pkg/credentials/aws/aws.go
@@ -23,9 +23,9 @@ import (
 )
 
 const (
-	awsAccountIDLength                    = 12
-	defaultSessionDuration                = 14400
-	defaultServiceUserCredentialsDuration = 24 * time.Hour
+	awsAccountIDLength                       = 12
+	defaultSessionDuration                   = 14400
+	defaultAWSServiceUserCredentialsDuration = 24 * time.Hour
 )
 
 // Credentials contains all data required for using AWS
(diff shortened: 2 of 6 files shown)


pkg/okctl/okctl.go
```diff
@@ -778,7 +781,7 @@ https://www.passwordstore.org/
 		github.NewAuthDeviceFlow(github.DefaultGithubOauthClientID, github.RequiredScopes()),
 	)
 
-	o.CredentialsProvider = credentials.New(aws.New(authStore, saml), gh)
+	o.CredentialsProvider = credentials.New(aws.New(authStore, envRetriever, samlRetriever), gh)
```
paulbes, line 784: Since `envRetriever` only "responds" to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` I think this will work. At least it decreases the likelihood that the user is surprised.

I know you aren't a huge fan of `NoInput`, which I understand. So I am  wondering if we should make the configuration of credentials more explicit in some other way? Perhaps we have a `--credentials-type` field with default set to `SAML` or something to that effect. Then we can add a second type `accesskey` or something? What do you think?

paulbes (commented):
Loving the tests ❤️

yngvark (approved):

docs/release_notes/0.0.39.md
```diff
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Allow for service user authentication through environment variables
```
yngvark, line 5: In the rest of the okclt code base we use the wording "service account". On the other hand, in those cases we're talking about kubernetes service accounts I think. Should we change to that for consistency?

pkg/credentials/aws/aws.go
```diff
@@ -200,6 +201,25 @@ func NewAuthStatic(creds *Credentials) *AuthStatic {
...
+type KeyGetter func(key string) (value string)
+
+// NewAuthEnvironment creates a retriever that fetches credentials from
+// environment variables
+func NewAuthEnvironment(region string, getter KeyGetter) Retriever {
+	credentials := &Credentials{
+		AccessKeyID:     getter("AWS_ACCESS_KEY_ID"),
+		SecretAccessKey: getter("AWS_SECRET_ACCESS_KEY"),
+		Region:          region,
+		Expires:         time.Now().Add(defaultServiceUserCredentialsDuration),
```
yngvark, line 214: Regarding "ServiceUser": See comment about consistency.

yngvark (approved):
See comments.

docs/release_notes/0.0.39.md
paulbes: I think this could be clarified, perhaps by simply calling it an `AWS service user`? They are different things at least, service account is a Kubernetes concept, while service user is an "us" concept. I guess?

deifyed: Argh. Explicit always win. Thanks for pointing it out

paulbes (approved):

PR #280, opened 2021-02-03
[KM142] Cloud formation stack outputs should also be exported, add DBSubnetGroupName output

paulbes:
Example output:
image

A cloud formation stack output cannot be referenced by another cloud formation stack unless it is exported. The DBSubnetGroupName is required when creating an RDS database.

**NOTE:** Please be aware that this comes with some side-effects (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html)

`After another stack imports an output value, you can't delete the stack that is exporting the output value or modify the exported output value. All of the imports must be removed before you can delete the exporting stack or modify the output value.`

and

`For each AWS account, Export names must be unique within a region.`

Closes: https://trello.com/c/I9rMXErA/142-legge-til-output-på-stacken-som-oppretter-databasesubnetgroup-om-å-eksportere-id-navn

## Description
- We now always export an output, so it can be referenced from another stack
- Added outputs to DBSubnetGroup

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.39.md
+++ b/docs/release_notes/0.0.39.md
@@ -1,6 +1,8 @@
-# Release 0.0.38
+# Release 0.0.39
 
 ## Features
+- Export all outputs, so they can be referenced by other cloud formation stacks
+- Add an output to the VPC stack for the DBSubnetGroupName
 
 ## Bugfixes
 34fe9be: Fix ASM permissions for external secrets GH#243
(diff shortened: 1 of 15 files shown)


deifyed (approved):

deifyed:
We should probably communicate the first caveat there. The second is irrelevant

yngvark (approved):

PR #289, opened 2021-02-04
feat: add --namespace, add deploy config

yngvark:
Replaces #196.

## Description

* Add `--namespace` argument to kubectl commands
* Add deployment configuration

## Types of changes

## Checklist:

### DIFF
--- a/userdocs/src/components/kubernetes.md
+++ b/userdocs/src/components/kubernetes.md
@@ -56,8 +56,8 @@ kubectl get secrets --namespace {your-namespace}
 kubectl describe secret postgres-config --namespace {your-namespace}
 ```
 
-In your `Deployment` add the following `env` configuration to let your pods get the `POSTGRES_PASSWORD` environment
-variable set on startup:
+If you have a postgres `Deployment`, add the following `env` configuration to let your pods get the `POSTGRES_PASSWORD`
+environment variable set on startup:
 
 ```yaml
 spec:
--- a/userdocs/src/components/kubernetes.md
+++ b/userdocs/src/components/kubernetes.md
@@ -2,7 +2,7 @@
 
 ## Kubernetes components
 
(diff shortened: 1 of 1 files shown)


deifyed (approved):
The `--namespace` flag isn't needed if one adds the Metadata.Namespace attribute. 

However, explicit is probably best. The users who knows, they know

userdocs/src/components/kubernetes.md
```diff
@@ -48,6 +49,29 @@ data:
...
 ```
 
+Inspect the secrets you just applied with:
+
+```bash
+kubectl get secrets --namespace {your-namespace}
+kubectl describe secret postgres-config --namespace {your-namespace}
+```
+
+If you have a postgres `Deployment`, add the following `env` configuration to let your pods get the `POSTGRES_PASSWORD`
```
paulbes, line 59: Should we explicitly state that this deployment needs to be applied within the same namespace as the secret above?

paulbes (commented):
Awesome stuff, one question

yngvark:
@deifyed That's what I was thinking as well, it doesn't hurt for those who know they can skip the parameter.

I think putting namespace in the yaml migh not be a good idea, as you can't easily apply the resource into whatever namespace you want. Or that's probably not entirely true, I guess kubectl --namespace overrides whatever the yaml says. But then you have to remember using --namespace. I'll just leave it as it.

userdocs/src/components/kubernetes.md
yngvark: Good idea, fixed.

deifyed:
Well. IMO applying a resource without `Metadata.Namespace` breaks the IAC principle. 

If you don't declare the infrastructure as it should be, you're expecting the user to know what namespace to put the resource in. 

Another argument could be that you want to reuse a resource, but again - that path diverges from IAC

@yngvark

yngvark:
Yes, good point, but I think it mostly applies if we're in an "IAC-setting". 

In a non-IAC setting I _do_ expect the user to know what namespace to put the resource in, but that is just how I personally work with k8s when I'm doing somthing with kubectl: I use `kubens` to set the active namespace, and then I want to apply stuff without having to override namespace all the time. 

But if the resource goes into a IAC-repository (and isn't a generic k8s-folder or helm chart for others to download), then it defintiely should be there.

I don't think it matters very much for this example though, as it's not the point of the article.

Issue #239, opened 2021-01-08
[FEATURE] Consider replacing AWSALBIngressController with AWSLoadBalancerController 


bsek:
We would like to use aws-load-balancer-type: "nlb-ip" which enables us to create network load balancers for use with pods running on Fargate nodes. This functionality is available in AWSLoadBalancerController but not in AWSALBIngressController.

As the new AWSLoadBalancerController is backwards compatible with the existing AWSALBIngressController, this should in theory not create any problems with the existing infrastructure but works as a simple replacement.

paulbes:
Absolutely, this is on the roadmap, more a question of when than if 👍

PR #295, opened 2021-02-05
[GH239] Replace ALB ingress controller with AWS load balancer controller

paulbes:
Replace ALB ingress controller with AWS load balancer controller (https://github.com/kubernetes-sigs/aws-load-balancer-controller)

## Description
This is a continuation of the "due diligence" process we have going, where we are moving towards the latest versions of our dependencies. 

AWS Load Balancer Controller is a controller to help manage Elastic Load Balancers for a Kubernetes cluster.

- It satisfies Kubernetes Ingress resources by provisioning Application Load Balancers.
- It satisfies Kubernetes Service resources by provisioning Network Load Balancers.

## Motivation and Context
The motivation is well described in #239 

## Types of changes

## Checklist:

### DIFF
--- a/pkg/api/core/service_serviceaccount.go
+++ b/pkg/api/core/service_serviceaccount.go
@@ -152,6 +152,54 @@ func (c *serviceAccount) CreateAlbIngressControllerServiceAccount(_ context.Cont
 	return account, nil
 }
 
+func (c *serviceAccount) CreateAWSLoadBalancerControllerServiceAccount(_ context.Context, opts api.CreateAWSLoadBalancerControllerServiceAccountOpts) (*api.ServiceAccount, error) {
+	err := opts.Validate()
+	if err != nil {
+		return nil, errInvalidInputs(err)
+	}
+
+	config, err := clusterconfig.NewAWSLoadBalancerControllerServiceAccount(
+		opts.ID.ClusterName,
+		opts.ID.Region,
+		opts.PolicyArn,
+		v1alpha1.PermissionsBoundaryARN(opts.ID.AWSAccountID),
+	)
+	if err != nil {
+		return nil, errBuildServiceAccount(err)
+	}
+
+	account, err := c.createServiceAccount(opts.CreateServiceAccountOpts, config)
+	if err != nil {
+		return nil, errCreateServiceAccount(err)
+	}
+
+	return account, nil
+}
+
+func (c *serviceAccount) DeleteAWSLoadBalancerControllerServiceAccount(_ context.Context, id api.ID) error {
+	err := id.Validate()
+	if err != nil {
+		return errInvalidInputs(err)
+	}
+
+	config, err := clusterconfig.NewAWSLoadBalancerControllerServiceAccount(
+		id.ClusterName,
+		id.Region,
+		"n/a",
+		v1alpha1.PermissionsBoundaryARN(id.AWSAccountID),
+	)
+	if err != nil {
+		return errBuildServiceAccount(err)
+	}
+
+	err = c.run.DeleteServiceAccount(config)
+	if err != nil {
+		return errDeleteServiceAccount(err)
+	}
+
+	return nil
+}
+
 func (c *serviceAccount) CreateExternalSecretsServiceAccount(_ context.Context, opts api.CreateExternalSecretsServiceAccountOpts) (*api.ServiceAccount, error) {
 	err := opts.Validate()
 	if err != nil {
--- a/pkg/api/core/service_serviceaccount.go
+++ b/pkg/api/core/service_serviceaccount.go
@@ -152,6 +152,7 @@ func (c *serviceAccount) CreateAlbIngressControllerServiceAccount(_ context.Cont
 	return account, nil
 }
 
+// nolint: lll
 func (c *serviceAccount) CreateAWSLoadBalancerControllerServiceAccount(_ context.Context, opts api.CreateAWSLoadBalancerControllerServiceAccountOpts) (*api.ServiceAccount, error) {
 	err := opts.Validate()
 	if err != nil {
--- a/pkg/api/core/store/noop/helm_noop.go
+++ b/pkg/api/core/store/noop/helm_noop.go
@@ -8,11 +8,11 @@ func (h *helmStore) SaveExternalSecretsHelmChart(helm *api.Helm) error {
 	return nil
 }
 
-func (h *helmStore) SaveAlbIngressControllerHelmChar(helm *api.Helm) error {
+func (h *helmStore) SaveAlbIngressControllerHelmChart(helm *api.Helm) error {
 	return nil
 }
 
-func (h *helmStore) SaveAWSLoadBalancerControllerHelmChar(helm *api.Helm) error {
+func (h *helmStore) SaveAWSLoadBalancerControllerHelmChart(helm *api.Helm) error {
 	return nil
 }
 
--- a/pkg/api/core/store/noop/helm_noop.go
+++ b/pkg/api/core/store/noop/helm_noop.go
@@ -12,6 +12,10 @@ func (h *helmStore) SaveAlbIngressControllerHelmChar(helm *api.Helm) error {
 	return nil
 }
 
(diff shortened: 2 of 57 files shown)


pkg/api/core/service_serviceaccount.go
```diff
@@ -152,6 +152,55 @@ func (c *serviceAccount) CreateAlbIngressControllerServiceAccount(_ context.Cont
...
+	return account, nil
+}
+
+func (c *serviceAccount) DeleteAWSLoadBalancerControllerServiceAccount(_ context.Context, id api.ID) error {
+	err := id.Validate()
+	if err != nil {
+		return errInvalidInputs(err)
+	}
+
+	config, err := clusterconfig.NewAWSLoadBalancerControllerServiceAccount(
```
deifyed, line 186: Why do we create a service account just to delete it (line 197)? Or does this function just create some kind of config despite its name

pkg/api/core/store/noop/helm_noop.go
```diff
@@ -12,6 +12,10 @@ func (h *helmStore) SaveAlbIngressControllerHelmChar(helm *api.Helm) error {
 	return nil
 }
 
+func (h *helmStore) SaveAWSLoadBalancerControllerHelmChar(helm *api.Helm) error {
```
deifyed, line 15: Missing 'T' at the end of function name.

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -243,11 +247,12 @@ func NewDefaultCluster(name, env, org, repo, team, accountID string) Cluster {
 			HighAvailability: true,
 		},
 		Integrations: &ClusterIntegrations{
-			ALBIngressController: true,
-			ExternalDNS:          true,
-			ExternalSecrets:      true,
-			Cognito:              true,
-			ArgoCD:               true,
+			ALBIngressController:      false,
+			AWSLoadBalancerController: true,
```
deifyed, line 251: Maybe we should get rid of the ALBIngressController? (Not in this PR) Or is it some kind of difference in functionality that we need to keep around

pkg/clusterconfig/clusterconfig.go
```diff
@@ -225,6 +225,22 @@ func NewAlbIngressControllerServiceAccount(clusterName, region, policyArn, permi
 	})
 }
 
+// NewAWSLoadBalancerControllerServiceAccount returns an initialised configuration
+// for creating an aws-load-balancer-controller service account
+func NewAWSLoadBalancerControllerServiceAccount(clusterName, region, policyArn, permissionsBoundaryArn string) (*v1alpha5.ClusterConfig, error) {
```
deifyed, line 230: Name is misleading. Should be something like `NewAWSLoadBalancerControllerServiceAccountConfig`

pkg/helm/charts/awslbc/awslbc.go
```diff
@@ -0,0 +1,196 @@
...
+	return &Values{
+		ReplicaCount:                  1, // nolint: gomnd
+		ClusterName:                   clusterName,
+		Region:                        region,
+		VpcID:                         vpcID,
+		TerminationGracePeriodSeconds: 10, // nolint: gomnd
+		EnableCertManager:             false,
+		ServiceAccount: ServiceAccount{
+			Create: false,
+			Name:   "aws-load-balancer-controller", // this is too fragile, refactor
```
deifyed, line 38: Refactor before or after merge?

pkg/helm/charts/awslbc/awslbc.go
```diff
@@ -0,0 +1,196 @@
...
+// NewDefaultValues returns the mapped values.yml containing the default
+// values
+func NewDefaultValues(clusterName, vpcID, region string) *Values {
+	return &Values{
+		ReplicaCount:                  1, // nolint: gomnd
+		ClusterName:                   clusterName,
+		Region:                        region,
+		VpcID:                         vpcID,
+		TerminationGracePeriodSeconds: 10, // nolint: gomnd
+		EnableCertManager:             false,
```
deifyed, line 35: What is this? Could we automate creating certificates for applications?

deifyed (changes requested):
8 code smells and a couple of comments

Good work though

yngvark:
@paulbes What do users with existing okctl clusters do to get this new AWS Load Balancer controller? I.e. how do they upgrade?

Should we write it in the user docs? And if it isn't possible to upgrade, still write something in the user docs aboutit.

deifyed:
@yngvark With `apply cluster` people should simply be able to set the old one to false and the new one to true 😍

pkg/api/core/service_serviceaccount.go
paulbes: I am not happy about this functionality, but essentially we are using `eksctl` to delete the service account for us. To do this, we need to provide a config the cluster name, region, service account name, etc.

If you look at the package name: `clusterconfig` you can see that your assumption is correct.

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
paulbes: Nope, we should remove it, just figured I would do it in a separate PR. 👍🏽

pkg/clusterconfig/clusterconfig.go
paulbes: Given the name of the package, I am not sure this is required? I even wonder if the linter might not give us a warning if we try to do that🤔


From: https://blog.golang.org/package-names
> Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.

pkg/helm/charts/awslbc/awslbc.go
paulbes: I believe it refers to this: https://github.com/jetstack/cert-manager, and yes, we could potentially use it now. It is essentially a service that uses lets encrypt to create certificates.

The main problem is that `AWS load balancer controller` and `cert-manager` kind of step on each others toes. With an ALB we typically terminate the encryption at the ALB and pass the request to the k8s pods in plaintext. Cert-manager attempts to terminate the encryption on the pod level. Now that we are using `AWS load balancer controller`, which supports Network Load Balancers (NLB). This might actually be possible.

☝🏽 this is based on my current understanding of the situation, without having done any further research.

pkg/helm/charts/awslbc/awslbc.go
paulbes: We are doing this many places already 😭 we should fix this, but perhaps as its own PR, so we can fix it at all places?

paulbes:
@yngvark short answer: they don't. However, you are touching on something very interesting:

- When do we start introducing upgrades?
- To properly manage expectations, what should we communicate to our users?

I think this is something we need to discuss within the team, and come to some common understanding about?

pkg/api/core/service_serviceaccount.go
deifyed: Ah. I thought clusterconfig was our code. My bad

pkg/clusterconfig/clusterconfig.go
deifyed: Might be. The function name makes me think it actually creates a service account. 

In any case, minor thing

pkg/helm/charts/awslbc/awslbc.go
deifyed: If its like how the nginx ingress controller + certbot stack works, it will terminate TLS in the ingress. In any case, its a nice improvement. More in line with the standard way of acquiring certificates with K8s. We can discuss this later

deifyed (approved):

paulbes:
This PR might be left hanging for a long time if we wait for the addons permissions boundary problem to be fixed in a released version of eksctl.

Downgrading to the last stable version (https://github.com/weaveworks/eksctl/releases/tag/0.32.0) without addons while we wait for a fixed release.

PR #306, opened 2021-02-10
✅ Add environment variable authentication for Github

deifyed:
## Description

Adds an option to use a token (PAT, Github Token) via an environment variable for authentication with Github

## Motivation and Context

It's the final piece needed for nightly testing of cluster creation.

Ref: [KM94](https://trello.com/c/k8LyFJLX/94-allow-for-programmatic-creation-of-clusters)

## Types of changes

## Checklist:

### DIFF
--- a/pkg/okctl/okctl.go
+++ b/pkg/okctl/okctl.go
@@ -845,10 +845,6 @@ https://www.passwordstore.org/
 
 // newCredentialsProvider knows how to load credentials
 func (o *Okctl) newCredentialsProvider() error {
-	if o.NoInput {
-		return errors.E(errors.Errorf("we only support retrieving credentials interactively for now"), errors.Invalid)
-	}
-
 	awsAuthenticator, err := o.getAWSAuthenticator()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/okctl/okctl.go
```diff
@@ -765,30 +758,72 @@ https://www.passwordstore.org/
...
-		},
+		&http.Client{Timeout: authenticatorTimeout},
 		github.NewAuthDeviceFlow(github.DefaultGithubOauthClientID, github.RequiredScopes()),
-	)
+	), nil
+}
+
+// newCredentialsProvider knows how to load credentials
+func (o *Okctl) newCredentialsProvider() error {
+	if o.NoInput {
```
paulbes, line 812: I guess you need to add something like: `&& github_device || aws_saml` for this if test to make sense?

paulbes (approved):
Only one thing that I think needs changing; otherwise, this is a beautiful PR 👏🏽

pkg/okctl/okctl.go
deifyed: Removed the o.NoInput flag. We can reintroduce it when its needed

PR #312, opened 2021-02-12
[KM89] Service quotas should ignore resources that are already provisioned 

paulbes:
Service quotas should ignore resources that are already provisioned 

## Description
If a cluster has already provisioned a VPC for itself, we should not write an error message to the user. The logic has also been rewritten so that we only write the status to the user if there are insufficient quotas available.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/create.go
+++ b/cmd/okctl/create.go
@@ -215,6 +215,9 @@ and database subnets.`,
 				return formatErr(err)
 			}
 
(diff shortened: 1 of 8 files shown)


deifyed (approved):
By order from PR author

yngvark (approved):
Nice!

ivaruf (approved):

PR #314, opened 2021-02-12
[KM108] Use explicit backends for keyring and build with CGO enabled

paulbes:
Use explicit backends for keyring and build with CGO enabled

## Description
The major part of this PR is fixing the issues with Keychain not being recognized on Darwin. When we build the Darwin binary with `CGO_ENABLED=0` we do not link in the `keychain` library, which means we have no way of interacting with that backend. With goreleaser we were building with CGO_ENABLED=0, therefore that backend was not being registered.

To make it work we need to build with `CGO_ENABLED=1`, this requires introducing a container for crosscompiling, since it requires certain build headers that aren't natively present on all developer machines.

- Pin to specific backends depending on os, for Darwin we use `Keychain` and for Linux we use `Pass`
- Check explicitly for the presence of pass on Linux and fail if it isn't present with a decent error message

## Types of changes

## Checklist:


deifyed (approved):
😍

ivaruf (approved):
Nice :+1:

yngvark (approved):
Great fix! Learned a lot by reading this PR.

PR #317, opened 2021-02-15
✅ Add emoji gitlog RFC

deifyed:
Not sure if its worth it to force any kind of git log system, I just thought I'd jot down what system I use to see what you guys think

### DIFF
--- a/docs/rfc/0003-emoji-gitlog.md
+++ b/docs/rfc/0003-emoji-gitlog.md
@@ -1,0 +1,70 @@
+# 0003. Emoji git log
+
+<!-- 
+    Date representing when the RFC was submittet for review 
+
+    Example:
+    Date: 08.02.2021
+-->
+Date: 15.02.2021
+
+## Context
+
+<!--
+    Short description of why this decision is needed
+
+    Example:
+    We want to better document our decisions
+-->
+
+Contributors should be able to easily create meaningful commit messages.
+
+## Decision
+
+<!--
+    Short summary of the decision
+
+    Example:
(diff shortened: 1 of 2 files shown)


paulbes (approved):
This is great, I personally use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), mainly because it helps me divide the commits into smaller chunks. I think this is the same.

I also agree that there isn't much value in enforcing a system, beyond commit messages that contain good information, until we do something useful with that system.

docs/rfc/0003-emoji-gitlog.md
```diff
@@ -0,0 +1,70 @@
...
+<!--
+    Short description of why this decision is needed
+
+    Example:
+    We want to better document our decisions
+-->
+
+Contributors should be able to easily create meaningful commit messages.
+
+## Decision
```
yngvark, line 22: This section describes the git format lag nicely, however it doesn't state what the actual decision is. Is this optional? Should we do it for some commits, most commits, or all?

I think the git log would look nice if we follow this RFC. However, it might be too intrusive to enforce it (require it from all/most commits), as it requires the developer to either set up some [helpers](https://github.com/ahmadawais/Emoji-Log#bashzsh-workflow) or somehow lookup emojis to do commits. So in my (current) opinion this should be optional.

yngvark (changes requested):

docs/rfc/0003-emoji-gitlog.md
deifyed: Yeah. I left that part out because I wanted some opinions. Mostly I just wanted to formalize my system and show it to y'all

deifyed: But yes, good point. I've yet to get a hang of writing RFC's

PR #318, opened 2021-02-15
[KM150] Improve user feedback when login goes wrong

paulbes:
Search for presence of error_message in the html and return if non-empty. Attempt to parse and validate the SAML assertion.

## Description
Instead of returning a cryptic error message to the user when they enter the wrong username/password or totp, we now extract the error message from the returned HTML and show that to the user:

- "Invalid username or password."
- "Invalid authenticator code."

Instead of returning relatively cryptic information about the SAML assertion, try to determine if user is capable of assuming the required role:

- "an error occurred when processing the SAML response from AWS:
you do not have permission to use the role:  arn:aws:iam::1234567890:role/oslokommune/iamadmin-SAM, ask for help in #kjørermiljø-support on slack"

## Types of changes

## Checklist:


deifyed (approved):
Nice!

yngvark (approved):
Looks good :+1:

PR #319, opened 2021-02-15
[KM27] Add cluster autoscaler to Kubernetes cluster

paulbes:
Add [cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to Kubernetes cluster

## Description
Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster when one of the following conditions is true:

- Some pods failed to run in the cluster due to insufficient resources.
- There are nodes in the cluster that have been underutilized for an extended period of time, and their pods can be placed on other existing nodes.

By introducing the cluster autoscaler, we can reduce costs during slow periods by scaling down the resource usage.

Each commit represents an isolated unit of code; thus, you can follow the way this is implemented by stepping through the commits in turn.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist:


deifyed (approved):
You're like the norwegian electricity infrastructure. Suffering from too much throughput

PR #323, opened 2021-02-16
[KM29] Add Amazon Elastic Block Store (EBS) CSI driver to cluster

paulbes:
Add [Amazon Elastic Block Store (EBS) CSI driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) to the Kubernetes cluster

Storage class declaration:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
volumeBindingMode: WaitForFirstConsumer
```

Persistent volume claim:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: block-claim
spec:
  accessModes:
    - ReadWriteOnce
  volumeMode: Block
  storageClassName: ebs-sc
  resources:
    requests:
      storage: 10Gi
```

Results in the following EBS volume:
image

## Description
This makes it possible to create `PersistentVolumeClaims` and attach them to pods.

**NB:** This does not work with pods scheduled onto fargate nodes

## Types of changes

## Checklist:


pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -196,6 +196,11 @@ type ClusterIntegrations struct {
 	// +optional
 	Autoscaler bool `json:"autoscaler,omitempty"`
 
+	// Blockstorage if set to true will install the EBS CSI block storage driver into the
+	// cluster, which makes it possible to create PersistentVolumeClaims in AWS
+	// +optional
```
deifyed, line 201: What does +optional actually do?

deifyed (approved):
Remember to update the changelog

yngvark (approved):
Looks good!

yngvark:
Does the user have to specify 

```yaml
  storageClassName: ebs-sc
```

in their PVC definition?

If yes - can we set it to default, so the user doesn't have to care about the underlying storage driver?

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
paulbes: It is based on k8s api machinery (if I remember correctly), essentially we could use it to autogenerate JSON schema, validation, etc.

paulbes:
Nice catch, yes we do, and yes we should make some or another storageclass a default. 👍🏽

deifyed (approved):

yngvark (approved):

PR #347, opened 2021-02-22
Add reconciliation requeue

yngvark:
## Description

Add rerun of reconciliatiors.

## Motivation and Context

ArgoCD reconciler fails sometimes due to timeout. In order to fix this we have add possibility to re run a reconciler if it fails.

https://trello.com/c/9QIqae53/176-handle-helm-k8s-timeouts-in-declarative-cluster

## Types of changes

## Checklist:

### DIFF
--- a/pkg/controller/reconciler/albingresscontroller_reconciler.go
+++ b/pkg/controller/reconciler/albingresscontroller_reconciler.go
@@ -30,29 +30,29 @@ func (z *albIngressReconciler) SetCommonMetadata(metadata *resourcetree.CommonMe
 }
 
 // Reconcile knows how to do what is necessary to ensure the desired state is achieved
-func (z *albIngressReconciler) Reconcile(node *resourcetree.ResourceNode) (*ReconcilationResult, error) {
+func (z *albIngressReconciler) Reconcile(node *resourcetree.ResourceNode) (result ReconcilationResult, err error) {
 	state, ok := node.ResourceState.(AlbIngressControllerResourceState)
 	if !ok {
-		return nil, errors.New("error casting ALB Ingress Controller state")
+		return result, errors.New("casting ALB Ingress Controller state")
 	}
 
 	switch node.State {
 	case resourcetree.ResourceNodeStatePresent:
-		_, err := z.client.CreateALBIngressController(z.commonMetadata.Ctx, client.CreateALBIngressControllerOpts{
+		_, err = z.client.CreateALBIngressController(z.commonMetadata.Ctx, client.CreateALBIngressControllerOpts{
 			ID:    z.commonMetadata.ClusterID,
 			VPCID: state.VpcID,
 		})
 		if err != nil {
-			return &ReconcilationResult{Requeue: true}, fmt.Errorf("error creating ALB Ingress controller: %w", err)
+			return result, fmt.Errorf("creating ALB Ingress controller: %w", err)
 		}
 	case resourcetree.ResourceNodeStateAbsent:
-		err := z.client.DeleteALBIngressController(z.commonMetadata.Ctx, z.commonMetadata.ClusterID)
+		err = z.client.DeleteALBIngressController(z.commonMetadata.Ctx, z.commonMetadata.ClusterID)
 		if err != nil {
-			return &ReconcilationResult{Requeue: true}, fmt.Errorf("error deleting ALB Ingress controller: %w", err)
+			return result, fmt.Errorf("deleting ALB Ingress controller: %w", err)
 		}
 	}
 
-	return &ReconcilationResult{Requeue: false}, nil
+	return result, nil
 }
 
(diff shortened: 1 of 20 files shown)


deifyed (approved):
Mob programmed, A-OK

docs/release_notes/0.0.42.md
```diff
@@ -0,0 +1,9 @@
+# Release 0.0.42
```
paulbes, line 1: 0.0.41 hasn't been released yet, would just add to that change log instead

pkg/controller/reconciler/albingresscontroller_reconciler.go
```diff
@@ -30,29 +30,29 @@ func (z *albIngressReconciler) SetCommonMetadata(metadata *resourcetree.CommonMe
 }
 
 // Reconcile knows how to do what is necessary to ensure the desired state is achieved
-func (z *albIngressReconciler) Reconcile(node *resourcetree.ResourceNode) (*ReconcilationResult, error) {
+func (z *albIngressReconciler) Reconcile(node *resourcetree.ResourceNode) (result ReconcilationResult, err error) {
```
paulbes, line 33: I would be a little bit careful with named return values, there is a section in effective-go on the matter: https://golang.org/doc/effective_go#named-results, where they name some situations where it can be useful:

> The names are not mandatory but they can make code shorter and clearer: they're documentation. If we name the results of nextInt it becomes obvious which returned int is which.

They also come with some problems though, such as variable shadowing: https://medium.com/@Nidhi11594/variable-shadowing-in-golang-f500e8e58931, our linter should pick this up, but beware.

In this situation, I am not sure that returning the empty `ReconciliationResult{}` is that much worse. Your mileage may vary, and to some extent, this is a personal preference. I know what I prefer 😬

pkg/controller/reconciler/argocd_reconciler.go
```diff
@@ -78,14 +78,26 @@ func (z *argocdReconciler) Reconcile(node *resourcetree.ResourceNode) (*Reconcil
 			AuthDomain:         resourceState.AuthDomain,
 			Repository:         repository,
 		})
+
 		if err != nil {
-			return &ReconcilationResult{Requeue: true}, fmt.Errorf("error creating argocd: %w", err)
+			// nolint: godox
+			// TODO: Need to identify the correct error
+			if strings.Contains(strings.ToLower(err.Error()), "timeout") {
+				fmt.Println(z.commonMetadata.Out, fmt.Errorf("got ArgoCD timeout: %w", err).Error())
```
paulbes, line 86: Should this be wrapped in a `if debug {}`?

pkg/controller/synchronize.go
```diff
@@ -81,16 +85,26 @@ func Synchronize(opts *SynchronizeOpts) error {
...
 
 // handleNode knows how to run Reconcile() on every node of a ResourceNode tree
-func handleNode(reconcilerManager reconciler.Reconciler, currentNode *resourcetree.ResourceNode) error {
-	_, err := reconcilerManager.Reconcile(currentNode)
-	if err != nil {
-		return fmt.Errorf("error reconciling node: %w", err)
+func handleNode(reconcilerManager reconciler.Reconciler, currentNode *resourcetree.ResourceNode) (err error) {
+	reconciliationResult := reconciler.ReconcilationResult{Requeue: true, RequeueAfter: 0 * time.Second}
+
+	for requeues := 0; reconciliationResult.Requeue; requeues++ {
```
paulbes, line 91: Maybe we should replace this with an exponential backoff approach: https://github.com/cenkalti/backoff/, the default values for this library might make sense to drop in as are: https://pkg.go.dev/github.com/cenkalti/backoff/v4#pkg-constants. Worth considering at least.

paulbes (commented):
This functionality makes a lot of sense 👍🏽

pkg/controller/reconciler/albingresscontroller_reconciler.go
deifyed: Yeah, this is my preference coming through. I just find it alot cleaner when I don't have to use `:` (when it makes sense). In the case of named results, I tend to explicitly declare variables to avoid shadowing

pkg/controller/reconciler/argocd_reconciler.go
deifyed: Could be, but without it we might get the correct error we want to check on later from a user. IDUNNO. But I'm off work and it would be nice to get this merged by tonights nightly :P

pkg/controller/synchronize.go
deifyed: I don't know anything about that package. I feel we get everything we need from the functionality implemented. Might want to add a multiplier at some point for exponential increase in delay time, but thats out of scope for this PR for now. Feel free to add it if you want, but I really want this merged before nightly tonight :P

PR #355, opened 2021-02-24
KM177: 🐛 Linux doesn't need pass

yngvark:
## Description
Don't require `pass` on Linux.

## Motivation and Context

#314 fixed so that macOS use it's native keychain instead of requiring pass. Unfortunately, now `okctl apply cluster` fails on Linux instead, as it now requires pass.

okctl should use the native keychain on Linux as well.

## How Has This Been Tested?

- [x] Tested on fresh install of ubuntu in virtualbox. Confirmed that I got the non-pass-keyring-popup asking for password.
- [x] Tested on our macOS instance in the cloud: Removed pass from it, and got same result as for linux, i.e. the macOS native keychain-thing.

For future reference, the way I tested it on macOS was to

* edit the okctl.go:getGithubAuthenticator method to contain

```
_ = k.Store("mytest", "hallo")
pw, err := k.Fetch("mytest")
```

and call `o.getGithubAuthenticator` from main.go:buildRootCommand as on one of the first lines.
* push it to a branch
* on the virtual mac, clone the branch, build it, and run it.
* run okctl apply cluster (or some other command)

All this so I could avoid logging in to my aws account (which the okctl apply cluster command need), which was very hard, as Nomachine running the virtual mac doesn't support inserting a clipboard value from the host (i.e. transferring account password and 2fa), and I don't really want to login to my account from that remote machine.

## Types of changes

## Checklist:

As the user documentation is currently being edited in another branch, I will update the docs related to this in that PR or a new PR.

### DIFF
--- a/pkg/keyring/keyring.go
+++ b/pkg/keyring/keyring.go
@@ -3,13 +3,7 @@ package keyring
 
 import (
 	"fmt"
-	"os"
-	"os/exec"
-	"path"
-	"path/filepath"
 	"runtime"
-
-	"github.com/mitchellh/go-homedir"
 
 	krPkg "github.com/99designs/keyring"
 	"github.com/oslokommune/okctl/pkg/config"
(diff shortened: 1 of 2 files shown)


deifyed (approved):
Looks neat.

pkg/keyring/keyring.go
```diff
@@ -57,49 +51,13 @@ func DefaultKeyringForOS() (krPkg.Keyring, error) {
 	case "darwin":
 		cfg.AllowedBackends = []krPkg.BackendType{krPkg.KeychainBackend}
 	case "linux":
-		cfg.AllowedBackends = []krPkg.BackendType{krPkg.PassBackend}
```
yngvark, line 60: Not specifying AllowedBackends means all backends are OK.

image

paulbes: Just beware of the words `all available`, as we saw with the keychain backend, this was not available because we were not building with CGO enabled, thus not satisfying the cgo build tag.

paulbes (approved):
I think this makes sense. The idea behind limiting the available backends was to control the "narrative", e.g., provide a useful error message if those two backends didn't work. I think we have more confidence in this now. Perhaps we should open up for available backends on mac also.

ivaruf (approved):

pkg/keyring/keyring.go
yngvark: I did manually testing on virtualbox+linux and a remote macOS instance to verify the keychain is working as expected.

PR #356, opened 2021-02-24
👌 Move NS record delegation test to after cluster

deifyed:
## Description

This PR creates a nameserver delegation tester reconciler, a reconciler that ensures that the nameservers has successfully been delegated. 

It also sets the identity provider as a dependency to the NS delegation test. 

## Motivation and Context

We can save some time by moving the actual test of a successful NS record delegation further down the process. That means that the `terraform apply` workflow and the actual dns propagation can do its thing while we set up more of the resources. 

## How Has This Been Tested?

Manually and things

## Types of changes

## Checklist:


yngvark (approved):
Approved in mob.

PR #362, opened 2021-02-25
[KM30] Add Loki for log retention and searching together with Promtail for scraping logs

paulbes:
Add Loki for log retention and searching together with Promtail for scraping logs

## Description
This got to be a lot larger than I had expected, mainly because it forced the creation of individual parts of a namespace. In addition, it made more sense to split out promtail and loki as separate entities, instead of using loki-stack directly. Should be easier to control what gets installed that way.

## Types of changes

## Checklist:


pkg/client/monitoring_client.go
```diff
@@ -0,0 +1,120 @@
...
+	SaveKubePromStack(stack *KubePromStack) (*store.Report, error)
+	RemoveKubePromStack(id api.ID) (*store.Report, error)
+	SaveLoki(loki *Loki) (*store.Report, error)
+	RemoveLoki(id api.ID) (*store.Report, error)
+	SavePromtail(Promtail *Promtail) (*store.Report, error)
+	RemovePromtail(id api.ID) (*store.Report, error)
+}
+
+// MonitoringState is a state layer implementation
+type MonitoringState interface {
```
deifyed, line 107: Don't we store any state regarding loki or promtail?

pkg/config/config.go
```diff
@@ -47,11 +47,13 @@ const (
 
 	DefaultEKSKubernetesVersion = "1.17"
 
-	DefaultChartApplyTimeout = 5 * time.Minute
+	DefaultChartApplyTimeout  = 5 * time.Minute
```
deifyed, line 50: 😍

pkg/helm/charts/loki/loki.go
```diff
@@ -0,0 +1,330 @@
...
+// New returns an initialised Helm chart for installing cluster-Loki
+func New(values *Values) *helm.Chart {
+	return &helm.Chart{
+		RepositoryName: "grafana",
+		RepositoryURL:  "https://grafana.github.io/helm-charts",
+		ReleaseName:    "loki",
+		Version:        "2.3.0",
+		Chart:          "loki",
+		Namespace:      "monitoring",
+		Timeout:        5 * time.Minute, // nolint: gomnd
```
deifyed, line 23: Shouldn't this use the DefaultHelmApply'amabobber variable?

pkg/helm/charts/promtail/promtail.go
```diff
@@ -0,0 +1,522 @@
...
+// New returns an initialised Helm chart for installing cluster-promtail
+func New(values *Values) *helm.Chart {
+	return &helm.Chart{
+		RepositoryName: "grafana",
+		RepositoryURL:  "https://grafana.github.io/helm-charts",
+		ReleaseName:    "promtail",
+		Version:        "3.1.0",
+		Chart:          "promtail",
+		Namespace:      "monitoring",
+		Timeout:        5 * time.Minute, // nolint: gomnd
```
deifyed, line 22: Shouldn't this use the DefaultHelmApplyChart'amabobber variable?

deifyed (approved):
Impressive work!

(Approved on the premise that you'll fix the broken tests)

pkg/kube/manifests/secret/secret.go
```diff
@@ -0,0 +1,98 @@
...
+		Name:      name,
+		Namespace: namespace,
+		Manifest:  manifest,
+		Ctx:       context.Background(),
+	}
+}
+
+// DeleteSecret deletes the secret
+func (s *Secret) DeleteSecret(_ kubernetes.Interface, config *rest.Config) (interface{}, error) {
+	client := kubernetes.NewForConfigOrDie(config)
```
deifyed, line 34: "Or die". I'm assuming this means a panic. As far as I can tell, we've been handling errors by returning them all the way back up the stack everywhere else. Why the change of behaviour?

pkg/client/core/service_monitoring_client.go
```diff
@@ -40,28 +50,206 @@ const (
...
+			"grafana_datasource": "1",
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// The datasources are only loaded during grafana startup, so we need
+	// to cycle grafana to have it pick up the changes
+	for _, replicas := range []int32{0, 1} {
```
yngvark, line 195: Can we be sure that the ordering is preserved? I.e. that what k8s actually performs is

```shell
kubectl scale --replicas=0
kubectl scale --replicas=1
```

and not vice versa?

I'm guessing that this most likely is not a problem, but I was unable to google an answer to this.

pkg/api/core/endpoint_kube.go
```diff
@@ -13,20 +13,44 @@ func makeCreateExternalDNSKubeDeploymentEndpoint(s api.KubeService) endpoint.End
...
 
-func makeCreateStorageClass(s api.KubeService) endpoint.Endpoint {
+func makeCreateNativeSecretEndpoint(s api.KubeService) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (response interface{}, err error) {
-		return s.CreateStorageClass(ctx, request.(api.CreateStorageClassOpts))
+		return s.CreateNativeSecret(ctx, request.(api.CreateNativeSecretOpts))
+	}
+}
+
+func makeDeleteNativeSecret(s api.KubeService) endpoint.Endpoint {
```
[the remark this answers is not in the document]
paulbes: Yeah, it is `v1.Secret`, as you have correctly surmised we don't really need this, so I will replace it.

Nice catch!

pkg/client/core/service_monitoring_client.go
yngvark, line 195: In loki_reconciler.go, we run CreateLoki every time, which means that grafana is scaled to 0 even though there might be no changes - is this correct? I'm guessing it's not worth the effort to make a better check here, but I'm mentioning in case you know of a super easy way of doing it.

pkg/api/core/run/kube_run.go
```diff
@@ -25,6 +29,76 @@ type kubeRun struct {
...
+		Description: fmt.Sprintf("scaling deployment: %s, at: %s, to: %d", opts.Name, opts.Namespace, opts.Replicas),
+	})
+	if err != nil {
+		return fmt.Errorf("creating secret: %w", err)
+	}
+
+	return nil
+}
+
+func (k *kubeRun) CreateNativeSecret(opts api.CreateNativeSecretOpts) (*api.NativeSecret, error) {
```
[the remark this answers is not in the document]
paulbes: We don't, changing to ConfigMap

pkg/client/core/store/filesystem/manifest.go
```diff
@@ -49,6 +89,15 @@ type ExternalSecret struct {
 	Manifests []string
 }
 
+func (s *manifestStore) RemoveExternalSecret(_ map[string]string) (*store.Report, error) {
+	// There is something fishy going on here, though I believe this will mostly be correct
```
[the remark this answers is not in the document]
paulbes: It is ugly basically, the whole Manifest thingy, should just be named `Kubectl` or something that makes more sense

pkg/client/monitoring_client.go
paulbes: We could probably find something to store, but for now I am not sure what information we have available that the user would need

pkg/kube/manifests/secret/secret.go
paulbes: Meltdown? I have no idea, we shouldn't be doing this, changing!

yngvark (approved):
Nice job 👍 Looking forward to check it out in my next cluster.

pkg/client/core/service_monitoring_client.go
paulbes: Great question, you can read about it here: https://golang.org/ref/spec#For_range
> For an array, pointer to array, or slice value a, the index iteration values are produced in increasing order, starting at element index 0.

So in this case, the order is given and will always be the same. For maps, however, the order is not given, and may be different for each iteration.

paulbes: I thought `CreateLoki` would only run if the current and expected trees were different, so if the `loki` directories aren't present it will run the check? If this is not the case, this isn't super pretty, I agree. I do believe that there is a Grafana API that can be called with the datasource instead, at least I have seen examples of this. I am not sure that it is worth the overhead given the minimal amount of impact this will have?

pkg/client/monitoring_client.go
deifyed: Ah no. No need to find something just to store something. Just wanted to make sure you didn't forget about state

pkg/client/core/service_monitoring_client.go
yngvark: > I thought CreateLoki would only run if the current and expected trees were different, so if the loki directories aren't present it will run the check?

That's right, I forgot about that.

No need to do anything here for now.

yngvark: Ah, this was more an k8s API question, not a Golang question. Do we know that the k8s api server will perform to operations from two rapidly incoming api calls in the order the api calls came?

paulbes: Aha, so they are blocking calls, so I hope so, but now you made me nervous 🤔 I will make a separate PR where we watch for the expected result before returning

deifyed: Very valid question

PR #371, opened 2021-03-05
[KM31] Add support for traces in Grafana with Tempo

paulbes:
Add support for traces in Grafana with Tempo

## Description
This adds [Tempo](https://grafana.com/oss/tempo/) to the cluster, which means that applications can push traces to tempo and then these can be queried through the Grafana UI.

I am also experimenting with a general create for Helm charts, which means we can reduce a lot of overhead.

## Types of changes

## Checklist:


pkg/api/core/endpoint_helm.go
```diff
@@ -55,14 +55,20 @@ func makeCreateLokiHelmChartEndpoint(s api.HelmService) endpoint.Endpoint {
...
-func makeDeleteHelmRelease(s api.HelmService) endpoint.Endpoint {
+func makeCreatePromtailHelmChartEndpoint(s api.HelmService) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (response interface{}, err error) {
-		return Empty{}, s.DeleteHelmRelease(ctx, request.(api.DeleteHelmReleaseOpts))
+		return s.CreatePromtailHelmChart(ctx, request.(api.CreatePromtailHelmChartOpts))
 	}
 }
 
-func makeCreatePromtailHelmChartEndpoint(s api.HelmService) endpoint.Endpoint {
+func makeCreateHelmRelease(s api.HelmService) endpoint.Endpoint {
```
deifyed, line 64: 😍

pkg/client/core/service_monitoring_client.go
```diff
@@ -51,13 +53,136 @@ const (
...
 }
 
+func (s *monitoringService) DeleteTempo(ctx context.Context, opts client.DeleteTempoOpts) error {
+	err := s.spinner.Start("Tempo")
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err = s.spinner.Stop()
```
deifyed, line 71: Should probably be `_ = s.spinner.Stop()` since we're not handling the error

pkg/kube/manifests/scale/scale.go
```diff
@@ -49,5 +55,33 @@ func (d *Scale) Scale(_ kubernetes.Interface, config *rest.Config) (interface{},
...
+	// desired state
+	t0 := time.Now()
+
+	ticker := time.NewTicker(5 * time.Second) // nolint: gomnd
+	defer ticker.Stop()
+
+	// I think this is a false positive, but probably not
+	// I am too tired to bother understanding it right now
+	// maybe some other day
+	// maybe never
```
deifyed, line 68: xD

README.md
```diff
@@ -114,8 +116,14 @@ We have a basic cluster up and running, and we are currently working on getting
 - [x] [Argo CD](https://github.com/argoproj/argo-cd) gives us continuous delivery
 - [x] [Kubernetes Prometheus](https://github.com/prometheus-operator/kube-prometheus) for monitoring
 - [x] [Amazon Elastic Block Store (EBS) CSI driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/) for block storage
-- [ ] [Loki](https://github.com/grafana/loki) for consuming logs
-- [ ] [Jaeger](https://www.jaegertracing.io/) for traces
+- [x] [Loki](https://github.com/grafana/loki) for consuming logs
+- [x] [Tempo](https://grafana.com/oss/tempo/) for processing trace
```
yngvark, line 120: trace -> traces? Or is it as intended?

deifyed (approved):
FLAWLESS VICTORY

pkg/helm/charts/tempo/tempo.go
```diff
@@ -0,0 +1,200 @@
...
+
+	err = tmpl.Execute(&buff, *v)
+	if err != nil {
+		return nil, err
+	}
+
+	return buff.Bytes(), nil
+}
+
+// nolint: lll
```
yngvark, line 59: Should we put a comment with an URL to the source of this values.yaml? Makes it a bit simpler when upgradring in the future (because values.yaml is tied to the tempo helm chart version), and if we for some other reason want to compare.

yngvark (approved):
Looks good 👍 A couple questions.

pkg/client/core/service_monitoring_client.go
paulbes: The idea, at some point in time, was to override the named return, but we should probably not be failing due to a failed spinner. So yes, agreed!

README.md
paulbes: Not intended, fixed!

pkg/helm/charts/tempo/tempo.go
paulbes: Yes! Very good idea ❤️

pkg/client/core/service_monitoring_client.go
deifyed: I was thinking more along the lines of if its `err =`, then a dev will have to look for when we do something depending on it. With `_ =` we dont have to look. Saves us approx 4 seconds.

PR #382, opened 2021-03-10
[NOTASK] Make it possible to define what users to add declaratively

paulbes:
Make it possible to define what users to add declaratively

## Description
We can now add a section as follows:

```
users:
- email: paul.beskow@origo.oslo.kommune.no
```

and when running `apply cluster` the identities will be added:

```
endpoint=create service=identitymanager/identitypool/identitypooluser
DEBU[0026] api.CreateIdentityPoolUserOpts{
  ID: api.ID{
    Region: "eu-west-1",
    AWSAccountID: "redacted",
    Environment: "paul",
    Repository: "okctl",
    ClusterName: "okctl-paul",
  },
  Email: "paul.beskow@origo.oslo.kommune.no",
  UserPoolID: "eu-west-FAKE",
}  endpoint=create service=identitymanager/identitypool/identitypooluser
... synchronizing: Users (elapsed: 31 seconds 4 milliseconds)INFO[0057] done with request, sending response           endpoint=create service=identitymanager/identitypool/identitypooluser
```

## Types of changes

## Checklist:


pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -231,11 +233,22 @@ type ClusterIntegrations struct {
 
 // Validate ensures there is no conflicting options
 func (c ClusterIntegrations) Validate() error {
-	if c.ArgoCD && !c.Cognito {
-		return errors.New("the identity provider cognito is required when using ArgoCD")
-	}
+	return validation.ValidateStruct(&c,
+		validation.Field(&c.Cognito, validation.When(c.ArgoCD, validation.Required.Error("is required when argocd is enabled"))),
```
deifyed, line 237: That looks cleaner than what I did :$

pkg/controller/reconciler/users_reconciler.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/oslokommune/okctl/pkg/apis/okctl.io/v1alpha1"
+
+	"github.com/oslokommune/okctl/pkg/client"
+	"github.com/oslokommune/okctl/pkg/controller/resourcetree"
+)
+
+// UsersState contains runtime data needed in Reconcile()
+type UsersState struct {
+	UserPoolID string
+	Users      []v1alpha1.ClusterUser
```
deifyed, line 15: This seems unused

pkg/controller/convert.go
```diff
@@ -52,6 +53,7 @@ func IdentifyResourcePresence(fs *afero.Afero, outputDir string, hzFetcher Hoste
 		hasArgoCD:                             directoryTester(fs, outputDir, config.DefaultArgoCDBaseDir),
 		hasDelegatedHostedZoneNameservers:     hz != nil && hz.IsDelegated,
 		hasDelegatedHostedZoneNameserversTest: false,
+		hasUsers:                              false, // this means we will always check? needs to be changed?
```
deifyed, line 56: This means it will always try to create the users in declaration.users. If the create users command is idempotent, this will work for now

pkg/controller/convert.go
```diff
@@ -82,6 +84,7 @@ func CreateResourceDependencyTree() (root *resourcetree.ResourceNode) {
 	identityProviderNode := createNode(delegatedNameserversConfirmedNode, resourcetree.ResourceNodeTypeIdentityManager)
 	kubePromStack := createNode(identityProviderNode, resourcetree.ResourceNodeTypeKubePromStack)
 	createNode(identityProviderNode, resourcetree.ResourceNodeTypeArgoCD)
```
deifyed, line 86: IMO: dependencies should reside beneath their parent. In other words, I'd suggest moving ArgoCD and Users up beneath identityprovider (line 85). I think this will improve readability

deifyed (approved):
Looks good. Couple of comments

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
paulbes: The only reason I changed it is because we can defer validation down the structs with `ozzo-validation`: https://github.com/go-ozzo/ozzo-validation#validatable-types, not that this is required here, but anyone reading the code will see that we always use the same pattern, a minor detail I know!

pkg/controller/reconciler/users_reconciler.go
paulbes: It is, I started on the delete side of the operations, and then saw the functionality was missing, and didn't think it was that important at this stage 🙈

pkg/controller/convert.go
paulbes: Good, that is what I thought, and it is idempotent 👍🏽

pkg/controller/reconciler/users_reconciler.go
deifyed: Ah good point

PR #389, opened 2021-03-12
[KM181] Manage the lifecycle of RDS Postgres (Part 2)

paulbes:
See #387 for more details.

**NB:** Do not merge until #390 has been merged into this PR

## Description
In this PR we implement the server and client sides and route up most of the things. We also create a hidden command for testing the setup. There will come more PRs, which will contain the reconciler for declaring the Postgres databases, a way of attaching to the database and a lambda for the rotation schedule. 

Preview of outputs:

```yaml
databases:
    mydb:
        type: postgres
        namespace: mydb
        endpointaddress: op1e2zk1kiflbdb.c5oiqv7tmmdx.eu-west-1.rds.amazonaws.com
        endpointport: 5432
        securitygroupid: sg-0a020b6f07f0a5c0e
        adminsecretname: mydb-postgres-admin
        databaseconfigmapname: mydb-postgres
```

```bash
$ kubectl describe cm mydb-postgres -n mydb
Name:         mydb-postgres
Namespace:    mydb
Labels:       <none>
Annotations:  <none>
Data
====
endpointAddress:
----
op1e2zk1kiflbdb.c5oiqv7tmmdx.eu-west-1.rds.amazonaws.com
endpointPort:
----
5432
Events:  <none>
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have updated the release notes (for the next release).

PR #387, opened 2021-03-12
[KM181] Manage the lifecycle of RDS Postgres

paulbes:
Manage the lifecycle of RDS Postgres

## Description
This is the first of a few PRs, with this PR we create the cloud formation template for an [RDS Postgres instance](https://aws.amazon.com/rds/postgresql/). In the following ones, we will setup the server and clientside integrations and add the setup into the declarative cluster.yaml file.

The cloud formation stack contains functionality for:

- Enabled automatic storage scaling
- Setting up a generated secrets manager secret with a rotation policy for the admin password
- Creating a security group that can be associated with a pod for gaining access to the RDS Postgres instance
- Creating the RDS Postgres instance itself

## Motivation and Context
The goal of okctl is to provide value to the end-users, we do this by simplifying the application lifecycle, which includes managing commonly used services such as Postgres

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] I have updated the release notes (for the next release).

PR #390, opened 2021-03-14
[KM181] Manage the lifecycle of RDS Postgres (Part 3)

paulbes:
See #387 and #389 for more details.

## Description
In this PR we create the lambda for the rotation schedule of the admin secret.

## Types of changes

## Checklist:


pkg/api/component_api.go
```diff
@@ -56,17 +56,60 @@ func (o *DeletePostgresDatabaseOpts) Validate() error {
...
+	)
+}
+
 // ComponentService defines a set of operations for creating
 // components that integrate with the Kubernetes cluster's
 // applications
 type ComponentService interface {
 	CreatePostgresDatabase(ctx context.Context, opts *CreatePostgresDatabaseOpts) (*PostgresDatabase, error)
 	DeletePostgresDatabase(ctx context.Context, opts *DeletePostgresDatabaseOpts) error
+	CreateS3Bucket(ctx context.Context, opts *CreateS3BucketOpts) (*S3Bucket, error)
```
ivaruf, line 104: What will we be using the s3 buckets for?

ivaruf (approved):

pkg/api/component_api.go
paulbes: Because of the `PermissionsBoundary` we have to create our own lambda function :( buuut, I actually think this might be a good thing long-term, because then we at least have support for creating lambdas from within okctl 🤷🏽 win some, loose some I guess

yngvark (approved):
:taco: :tada:

pkg/static/rotater/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:20.04
+
+COPY . /source
+
+RUN apt-get update
+RUN apt-get -y install git zip
+
+# https://github.com/jkehler/awslambda-psycopg2
+RUN git clone https://github.com/jkehler/awslambda-psycopg2.git
```
deifyed, line 9: You should probably lock the clone to a certain commit to avoid a potential remote code execution problem.

Also, we should switch to a less privileged user with
```
USER app_user
```

pkg/static/rotater/rotater.go
```diff
@@ -0,0 +1,10 @@
+// Package rotater embeds the lambda function code zip file
+package rotater
+
+import _ "embed"
+
+// LambdaFunctionZip contains the embedded data for the
+// rotater lambda
+// nolint: gochecknoglobals
+//go:embed lambda_function.zip
```
deifyed, line 9: 😍

deifyed (approved):

cmd/okctl/delete_postgres.go
```diff
@@ -0,0 +1,85 @@
...
+		validation.Field(&o.ID, validation.Required),
+		validation.Field(&o.ApplicationName, validation.Required),
+	)
+}
+
+// nolint: funlen
+func buildDeletePostgresCommand(o *okctl.Okctl) *cobra.Command {
+	opts := &deletePostgresOpts{}
+
+	cmd := &cobra.Command{
```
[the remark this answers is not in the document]
paulbes: This is just temporary, just to have something that is easier to test with

pkg/static/rotater/Dockerfile
paulbes: Agreed on the commit, is the user thing necessary when we just use the dockerfile to build the zip?

deifyed: Hmm. No clue. Just know its a common hardening technique

pkg/static/rotater/lambda_function.py
```diff
@@ -1,18 +1,11 @@
-// Package static contains some static data, we could probably have
-// used embed here, but maybe some other time..
-package static
-
-// SecretsManagerPostgresRotationSingleUserBody contains the code for the AWS Secrets Manager rotation for postgres
-// - https://github.com/aws-samples/aws-secrets-manager-rotation-lambdas/blob/master/SecretsManagerRDSPostgreSQLRotationSingleUser
```
yngvark: Can we keep this comment, so we know where to get the latest content for this file?

paulbes: good point, will add it to the readme

PR #393, opened 2021-03-16
KM184 👌 Clean up cluster declaration

deifyed:
## Description

* Adds documentation to `scaffold cluster` template
* Renames primaryHostedZone parent domain to clusterRootURL
* Communicates optional values
* Ensures optional values are optional
* Moves constants to `constant` package

## Motivation and Context

[KM184](https://trello.com/c/FbKfSTEy/184-clean-up-clusteryaml)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


.github/workflows/nightly.yaml
```diff
@@ -29,30 +29,31 @@ jobs:
...
-            integrations:
-              argoCD: true
-              awsLoadBalancerController: true
-              cognito: true
-              externalDNS: true
-              externalSecrets: true
-            vpc:
-              cidr: 192.168.0.0/20
-              highAvailability: true
+          apiVersion: okctl.io/v1alpha2
```
yngvark, line 32: We are using v1alpha1 other places.

yngvark (changes requested):

cmd/okctl/applycluster.go
```diff
@@ -78,11 +76,16 @@ func buildApplyClusterCommand(o *okctl.Okctl) *cobra.Command {
 			o.AWSCredentialsType = opts.AWSCredentialsType
 			o.GithubCredentialsType = opts.GithubCredentialsType
 
-			opts.Declaration, err = inferClusterFromStdinOrFile(o.In, opts.File)
+			opts.Declaration, err = commands.InferClusterFromStdinOrFile(o.In, opts.File)
```
paulbes, line 79: Hell, yeah! ❤️

cmd/okctl/create.go
```diff
@@ -5,6 +5,8 @@ import (
 	"path"
 	"regexp"
 
+	"github.com/oslokommune/okctl/pkg/config/constant"
```
paulbes, line 8: When I see it like this, I get it! It flows nicely ❤️

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -25,14 +27,14 @@ type Cluster struct {
...
 	// this cluster will integrate with.
 	Github ClusterGithub `json:"github"`
 
-	// PrimaryDNSZone defines the main primary zone to associate with this
-	// cluster. This will be the zone that we will use to create domains
+	// ClusterRootURL defines the main primary zone to associate with this
+	// cluster. This will be the zone that we will use to create subdomains
 	// for auth, ArgoCD, etc.
-	PrimaryDNSZone ClusterDNSZone `json:"primaryDNSZone"`
+	ClusterRootURL string `json:"clusterRootURL"`
```
paulbes, line 37: A [url](https://en.wikipedia.org/wiki/URL) is a very specific thing, so this is misleading. `PrimaryDNSZone` isn't much better though.. 🤔 

Maybe something like:

`BaseDomain`: also slightly misleading, since it strictly only includes domain and TLD
`SubDomain`: probably the most accurate, but maybe not what we want to communicate
`RootDomain`: kind of combines some of these

In reality, we don't really require it to be a subdomain, if someone wants to create their own domain, they are free to do so.

Maybe we should just call it:

`DefaultClusterDomain`: the default domain used for all kinds of grafana setups and stuff. Also leaves room for multiple domains in the future if someone would like to do so

I don't know, but I don't think we should have `URL` in there at leaast

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -270,18 +267,14 @@ func NewDefaultCluster(name, env, org, repo, team, accountID string) Cluster {
 			Region:      "eu-west-1",
 			AccountID:   accountID,
 		},
-		PrimaryDNSZone: ClusterDNSZone{
-			ParentDomain:  fmt.Sprintf("%s-%s.oslo.systems", name, env),
-			ReuseExisting: false,
-		},
+		ClusterRootURL: fmt.Sprintf("%s-%s.oslo.systems", name, env),
```
paulbes, line 270: This is basically the `ClusterName`, which we use other places

pkg/commands/scaffoldcluster.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+const clusterTemplate = `apiVersion: okctl.io/v1alpha1
+kind: Cluster
+
+# For help finding values, see https://okctl.io/getting-started/create-cluster
+metadata:
+  # Account ID is your AWS account ID
+  accountID: '{{ .AWSAccountID }}'
+  # Environment is the name you use to identify the type of cluster it is. Common names are production, test, staging
```
paulbes, line 56: Lets put in a whitespace between each Comment:Variable pair? Will make it a lot easier to read

pkg/commands/scaffoldcluster.go
```diff
@@ -0,0 +1,117 @@
...
+		Name:            strings.ToLower(opts.Name),
+		AWSAccountID:    opts.AWSAccountID,
+		Environment:     strings.ToLower(opts.Environment),
+		Organization:    opts.Organization,
+		RepositoryName:  opts.RepositoryName,
+		OutputDirectory: opts.OutputDirectory,
+	}
+}
+
+const clusterTemplate = `apiVersion: okctl.io/v1alpha1
```
paulbes, line 49: This is such an improvement!

paulbes (changes requested):
Really nice upgrade of the cluster.yaml!

I think we need to change the clusterRootURL name though, and I am willing to fight for it 😬

cmd/okctl/create.go
deifyed: woooh im glad!

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
yngvark: Agree, URL is incorrect. Domain makes sense. I'll vote for `RootDomain` for its simplicity. `DefaultClusterDomain` could work too, but I think it raises more questions than being explanatory. But maybe with a good comment it can work.

pkg/commands/scaffoldcluster.go
deifyed: @paulbes do you mean newline?

paulbes: Yeah!

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
yngvark: We went for `ClusterRootDomain`, as it was hopefully more descriptive than just `RootDomain`.

pkg/commands/scaffoldcluster.go
yngvark: Done

pkg/commands/scaffoldcluster.go
yngvark: :pray:

paulbes (approved):

yngvark (approved):

PR #397, opened 2021-03-17
KM196 🐛 Fix error due to ExternalDNS timing out

deifyed:
## Description

- Identifies a timeout as early as possible in the `kubectl` functionality
- Ensures the timeout error travels from the server to the client reconciler  

## Motivation and Context

[KM196](https://trello.com/c/7yhX8th6/193-setup-requeue-for-waiterrwaittimeout-in-externaldns-reconciler-create)

## How Has This Been Tested?

Added a failure point in a server side call for the first reconciler to run (primary hosted zone) and made sure I could identify the same type of error in the reconciler that called it. 

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.49.md
+++ b/docs/release_notes/0.0.49.md
@@ -3,5 +3,6 @@
 ## Features
 
 ## Bugfixes
+KM196: Add handling of timeout errors for ExternalDNS (#397)
 
 ## Other
--- a/pkg/client/core/api/rest/client.go
+++ b/pkg/client/core/api/rest/client.go
@@ -9,6 +9,10 @@ import (
 	"net/http"
 	"strings"
 
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+
+	"github.com/mishudark/errors"
+
 	"github.com/sanity-io/litter"
 )
 
@@ -41,6 +45,7 @@ func (c *HTTPClient) DoDelete(endpoint string, body interface{}) error {
 }
 
 // Do performs the request
+// nolint: funlen
 func (c *HTTPClient) Do(method, endpoint string, body interface{}, into interface{}) error {
 	if c.Debug {
 		_, err := fmt.Fprintf(c.Progress, "client (method: %s, endpoint: %s) starting request: %s", method, endpoint, litter.Sdump(body))
@@ -72,7 +77,7 @@ func (c *HTTPClient) Do(method, endpoint string, body interface{}, into interfac
 	}
 
 	if resp.StatusCode >= 400 { // nolint: gomnd
-		return fmt.Errorf("request failed with %s, because: %s", http.StatusText(resp.StatusCode), string(out))
+		return deserializeErrorPayload(out)
 	}
 
 	defer func() {
@@ -103,4 +108,38 @@ func (c *HTTPClient) Do(method, endpoint string, body interface{}, into interfac
 
 func pretty(msg, method, endpoint string) string {
 	return fmt.Sprintf("%s: %s, %s", msg, method, endpoint)
+}
+
+type serializedError struct {
+	Detail map[string]interface{} `json:"detail,omitempty"`
+	Type   string                 `json:"type"`
+	Error  string                 `json:"error"`
+	Code   errors.Kind            `json:"code"`
+}
+
+func (s serializedError) Validate() error {
+	return validation.ValidateStruct(&s,
+		validation.Field(&s.Error, validation.Required),
+	)
+}
+
+// deserializeErrorPayload converts an JSON error payload from the backend to a structured error
+func deserializeErrorPayload(jsonContent []byte) error {
+	data := serializedError{}
+
+	err := json.Unmarshal(jsonContent, &data)
+	if err != nil {
+		return errors.E(
+			fmt.Errorf("unmarshalling error from server side: %w", err),
+			errors.Internal)
+	}
+
+	if err = data.Validate(); err != nil {
+		return errors.E(
+			fmt.Errorf("validating deserialized error with content %s: %w", string(jsonContent), err),
+			errors.Unmarshal,
+		)
+	}
+
+	return errors.E(errors.New(data.Error), data.Code, data.Detail)
 }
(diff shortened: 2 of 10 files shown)


paulbes (commented):
Based on our earlier discussion, this is kind of what I was thinking:

```go
package pkg

import (
	"context"
	"encoding/json"
	"fmt"
	"net/http/httptest"
	"testing"

	"github.com/stretchr/testify/assert"

	"github.com/go-kit/kit/transport/http"
	"github.com/mishudark/errors"
)

type ErrTimeout struct {
	Message string
}

func (e *ErrTimeout) Error() string {
	return e.Message
}

type Error struct {
	Type   string  `json:"type"`
	Error  string  `json:"error"`
	Code   int     `json:"code"`
	Detail *Detail `json:"detail,omitempty"`
}

type Detail struct {
	Timeout *ErrTimeout `json:"timeout,omitempty"`
}

func Something() interface{} {
	recorder := httptest.NewRecorder()

	http.DefaultErrorEncoder(
		context.Background(),
		errors.E(
			fmt.Errorf("this is what went wrong"),
			"oh no",
			map[string]interface{}{
				"timeout": &ErrTimeout{Message: "because time ran ouuuuuuuuuut"},
			},
			errors.Transient,
		),
		recorder,
	)

	switch recorder.Result().StatusCode {
	case errors.Transient.StatusCode():
		to := &Error{}
		_ = json.NewDecoder(recorder.Result().Body).Decode(to)
		return to
	default:
		return fmt.Errorf("meh")
	}
}

func TestSomething(t *testing.T) {
	got := Something()
	assert.Equal(t, &Error{
		Type:  "transient error",
		Error: "oh no",
		Code:  10,
		Detail: &Detail{
			Timeout: &ErrTimeout{
				Message: "because time ran ouuuuuuuuuut",
			},
		},
	}, got)
}
```

pkg/client/core/api/rest/client.go
```diff
@@ -104,3 +116,22 @@ func (c *HTTPClient) Do(method, endpoint string, body interface{}, into interfac
...
+	Code   errors.Kind            `json:"code"`
+}
+
+// jsonAsMishudarkError converts an error from the backend to a mishudark error
+func jsonAsMishudarkError(jsonContent []byte, defaultError error) error {
+	data := serializedError{}
+
+	err := json.Unmarshal(jsonContent, &data)
+	if err != nil {
+		return defaultError
```
paulbes, line 133: The go JSON marshaller isn't strict, so if something doesn't match the struct you pass in, it will not fail. It will generally only fail if we are not unmarshalling proper JSON (https://stackoverflow.com/questions/19633763/unmarshaling-json-in-golang-required-field)

pkg/client/core/api/rest/client.go
```diff
@@ -104,3 +116,22 @@ func (c *HTTPClient) Do(method, endpoint string, body interface{}, into interfac
...
+
+type serializedError struct {
+	Detail map[string]interface{} `json:"detail,omitempty"`
+	Type   string                 `json:"type"`
+	Error  string                 `json:"error"`
+	Code   errors.Kind            `json:"code"`
+}
+
+// jsonAsMishudarkError converts an error from the backend to a mishudark error
+func jsonAsMishudarkError(jsonContent []byte, defaultError error) error {
```
paulbes, line 128: We shouldn't have `Mishudark` in the name of this function

paulbes (changes requested):

pkg/client/core/api/rest/client.go
deifyed: So in other words, create an internal error instead only complaining about unmarshalling being given bad content?

paulbes: In the example I posted earlier I always decode into the structure below if there is an error. Basically, we need to guarantee that any error that is returned by the server conforms to some "schema" and then we need to unmarshal into that schema

```go
type ErrTimeout struct {
	Message string
}

func (e *ErrTimeout) Error() string {
	return e.Message
}

type Error struct {
	Type   string  `json:"type"`
	Error  string  `json:"error"`
	Code   int     `json:"code"`
	Detail *Detail `json:"detail,omitempty"`
}

type Detail struct {
	Timeout *ErrTimeout `json:"timeout,omitempty"`
}
```

docs/release_notes/0.0.49.md
```diff
@@ -3,5 +3,6 @@
 ## Features
 
 ## Bugfixes
+KM196: Add handling of timeout errors for ExternalDNS
```
yngvark, line 6: Append `(#397)` to the line?

pkg/client/core/api/rest/client.go
paulbes: I misunderstood your code! Just disregard me

paulbes (approved):

pkg/client/core/api/rest/client_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+			withKind: errors.Timeout,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			server := httptest.NewServer(testMarshalledErrorServer{tc.withKind})
```
yngvark, line 46: From NewServer doc:
```
The caller should call Close when finished, to shut it down.
```

pkg/controller/synchronize.go
```diff
@@ -64,18 +63,19 @@ func Synchronize(opts *SynchronizeOpts) error {
...
 		if requeues == constant.DefaultMaxReconciliationRequeues {
-			return errors.New("maximum allowed reconciliation requeues reached")
+			return fmt.Errorf("maximum allowed reconciliation requeues reached: %w", err)
 		}
 
 		time.Sleep(reconciliationResult.RequeueAfter)
 
 		reconciliationResult, err = reconcilerManager.Reconcile(currentNode)
-		if err != nil {
+		if err != nil && !reconciliationResult.Requeue {
```
yngvark, line 78: Is this something we should test?

Though if it is, maybe it's outside the scope of this PR.

yngvark (approved):
Splendid!

PR #410, opened 2021-03-22
✅ Add requeue to identitymanager upon timeout

deifyed:
## Description

Adds a catcher for stack creation timeouts, and ensures the identitymanager's reconciler requeues upon a timeout type error

## Motivation and Context

[KM203](https://trello.com/c/KYvl0GiF)

## Types of changes

## Checklist:

### DIFF
--- a/pkg/cfn/runner_test.go
+++ b/pkg/cfn/runner_test.go
@@ -1,6 +1,7 @@
 package cfn_test
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/service/cloudformation"
@@ -34,7 +35,7 @@ func TestCreate(t *testing.T) {
 					DescribeStacksResponse(cloudformation.StackStatusCreateFailed).
 					DescribeStackEventsSuccess(),
 			),
-			expect:      "stack: something, failed events: ec2: something went wrong",
+			expect:      `{"detail":{"failed events":["ec2: something went wrong"]},"type":"internal error","error":"stack failed","code":7}`,
 			expectError: true,
 		},
 	}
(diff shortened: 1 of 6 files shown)


pkg/cfn/runner_test.go
```diff
@@ -44,8 +45,10 @@ func TestCreate(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			err := tc.runner.CreateIfNotExists(mock.DefaultStackName, []byte{}, nil, 10)
 			if tc.expectError {
+				result, _ := json.Marshal(err)
```
yngvark, line 48: You perhaps explained this before, but why are we converting to json, can we not just create an expected error struct with Mishudark?

pkg/controller/reconciler/identitymanager_reconciler.go
```diff
@@ -56,6 +56,8 @@ func (z *identityManagerReconciler) Reconcile(node *resourcetree.ResourceNode) (
 			HostedZoneID: resourceState.HostedZoneID,
 		})
 		if err != nil {
+			result.Requeue = errors.IsKind(err, errors.Timeout)
```
yngvark, line 59: This can create an eternal loop if there are endless of timeout errors.  Perhaps it can happen if we for instance have too low timeout, but there might be other unknown reason that leads to a timeout error.

Just mentioning, if you think it is safe enough, leave it. If not, we need a maximum retries thingy here.

yngvark (approved):

pkg/controller/reconciler/identitymanager_reconciler.go
deifyed: The maximum retries thingy happens [here](https://github.com/oslokommune/okctl/blob/f4fa56da142ac97ff9f77d1b2c397fe8082e39be/pkg/controller/synchronize.go#L71). This is just a signal indicating what the reconciler wants to do.

pkg/cfn/runner_test.go
deifyed: Might work. I can give it another try. Reason why I did this is because I'm not proficient enough in Go to sufficiently compare the two.

The reason why I converted it to JSON is because that way I could see all the information the mishudark error contained.

pkg/controller/reconciler/identitymanager_reconciler.go
yngvark: Ooh right, totally forgot that. Good then :+1:

pkg/cfn/runner_test.go
yngvark: I see, that makes sense. :+1:

paulbes (approved):
I think you are right @deifyed, we need to consider a different strategy for retrying on timeouts 🤔

PR #436, opened 2021-03-30
👌 Truncate log lines to 5000 bytes

yngvark:
## Description
It's quite hard to read the log from the nightly these days, it is 87MB. I found single log line being 64MB!

## Motivation and Context

I want to read the log using a regular editor or web gui, not having to resort to shell truncating magic to make sense of it.

[ ] TODO: Add release notes

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] I have updated the release notes (for the next release).

PR #454, opened 2021-04-09
🐛 Fix truncate log not working

yngvark:
## Description

#436 and #444 introduced a fatal bug where okctl apply cluster doesn't work anymore. Whenever the log is truncated, okctl aborts with error

```
Error: synchronizing declaration with state: reconciling node: creating vpc: failed to parse response: illegal base64 data at input byte 4805
```

Not sure why, but logging anything but base64 characters results in this error, so the solution is to only log base64-characters.

## How Has This Been Tested?

I verified that this fix ran OK on the nightly github action workflow, and with apply cluster locally.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/truncate/truncate.go
+++ b/pkg/truncate/truncate.go
@@ -3,8 +3,14 @@ package truncate
 
 import "fmt"
 
+const truncateFormat = "XXXtruncated%dbytesXXX"
+
 // String truncates a string to the minimum of its length and the given max length
 func String(s *string, maxLength int) string {
+	if s == nil {
+		return ""
+	}
+
 	// This way of doing substrings assumes ASCII and doesn't support UTF-8.
 	// See https://stackoverflow.com/a/56129336
 	truncateLength := min(maxLength, len(*s))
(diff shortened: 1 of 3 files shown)


pkg/truncate/truncate.go
```diff
@@ -3,30 +3,40 @@ package truncate
...
-		truncated += fmt.Sprintf(" [truncated %d bytes]", bytesTruncated)
+		truncated += fmt.Sprintf(truncateFormat, bytesTruncated)
 	}
 
 	return truncated
 }
 
 // Bytes truncates a byte array to the minimum of its length and the given max length
 func Bytes(b []byte, maxLength int) []byte {
+	if b == nil {
```
paulbes, line 29: Nice!

pkg/truncate/truncate.go
```diff
@@ -3,30 +3,40 @@ package truncate
 
 import "fmt"
 
+const truncateFormat = "XXXtruncated%dbytesXXX"
```
paulbes, line 6: Did we ever understand why this was required?

paulbes (approved):
Great stuff! ❤️

pkg/truncate/truncate.go
yngvark: Unfortunately not. I had a look on the surrounding code without understanding it, and decided not to spend more time on it.

PR #456, opened 2021-04-12
KM230 👌 Convert apply application to reconciliation pattern

deifyed:
## Description

Makes `apply application` use the reconciler/operator pattern as in `apply cluster`

## Motivation and Context

[KM230](https://trello.com/c/xbYLxDEe)

Julius likes this pattern. (Also it provides a more "natural" way to introduce more features to an application, f.eks ECR)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/apply_application.go
+++ b/cmd/okctl/apply_application.go
@@ -8,7 +8,6 @@ import (
 	"github.com/oslokommune/okctl/pkg/apis/okctl.io/v1alpha1"
 	"github.com/oslokommune/okctl/pkg/client"
 	"github.com/oslokommune/okctl/pkg/commands"
-	"github.com/oslokommune/okctl/pkg/config/state"
 	"github.com/oslokommune/okctl/pkg/controller"
 	"github.com/oslokommune/okctl/pkg/controller/reconciler"
 	"github.com/oslokommune/okctl/pkg/controller/resourcetree"
@@ -95,7 +94,7 @@ func buildApplyApplicationCommand(o *okctl.Okctl) *cobra.Command {
 				// We should pass inn cluster fetched from state here when state rewrite is done
 				Declaration: &v1alpha1.Cluster{
 					Github: v1alpha1.ClusterGithub{
-						Repository: getFirstGithubRepositoryURL(o.RepoStateWithEnv.GetGithub().Repositories),
+						Repository: commands.GetFirstGithubRepositoryURL(o.RepoStateWithEnv.GetGithub().Repositories),
 						OutputPath: o.RepoStateWithEnv.GetMetadata().OutputDir,
 					},
 				},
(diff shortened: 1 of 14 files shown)


cmd/okctl/apply_application.go
```diff
@@ -91,15 +80,49 @@ func buildApplyApplicationCommand(o *okctl.Okctl) *cobra.Command {
...
-			err = services.ApplicationService.ScaffoldApplication(o.Ctx, scaffoldOpts)
+			reconciliationManager := reconciler.NewCompositeReconciler(spin,
+				reconciler.NewApplicationReconciler(services.ApplicationService),
+			)
+
+			reconciliationManager.SetCommonMetadata(&resourcetree.CommonMetadata{
+				Ctx:       o.Ctx,
+				Out:       o.Out,
+				ClusterID: *scaffoldOpts.ID,
+				// We should pass inn cluster fetched from state here when state rewrite is done
```
ivaruf, line 94: Add TODO? Actually valid case :)

ivaruf (approved):
Looks great :+1:

pkg/controller/reconciler/application_reconciler.go
```diff
@@ -0,0 +1,66 @@
...
+			HostedZoneID:     state.PrimaryHostedZoneID,
+			HostedZoneDomain: state.PrimaryHostedZoneDomain,
+			IACRepoURL:       a.commonMetadata.Declaration.Github.Repository,
+			Application:      state.Declaration,
+		})
+		if err != nil {
+			return ReconcilationResult{}, err
+		}
+	case resourcetree.ResourceNodeStateAbsent:
+		return ReconcilationResult{}, errors.New("deletion of applications is not implemented")
```
yngvark, line 55: Should we add this as a Trello-task?

pkg/controller/reconciler/application_reconciler.go
```diff
@@ -0,0 +1,66 @@
+package reconciler
```
yngvark, line 1: Have you considered putting this reconciler in a separate directory? E.g. `pkg/controller/reconciler/application/`, next to `pkg/controller/reconciler/cluster/` for the others?

yngvark (approved):
Looks good :+1:

pkg/controller/reconciler/application_reconciler.go
deifyed: Yeah, but couldn't really make up my mind. The certificate reconciler is for example used by both the cluster and applications. Meaning we'd need a `common`, `cluster` and an `application`.

Do chime in if you have any thoughts

pkg/controller/reconciler/application_reconciler.go
deifyed: Good idea

pkg/controller/reconciler/application_reconciler.go
yngvark: Right, I knew there was going to be some issues.

Slightly related: In convert.go we have both `CreateResourceDependencyTree` and `CreateApplicationResourceDependencyTree`, so we don't have any clear separation there either. So all together I think we can leave it as it is, it's simple enough.

deifyed: Also a good point. 

I kinda settled on waiting until we got one more case for creating dependency trees / reconcilers to see if there was an obvious pattern

PR #464, opened 2021-04-14
[NOTASK] Establish a janitor that helps us keep things under control

paulbes:
Establish a janitor that helps us keep things under control

## Description
Over time we build up trash, what better way to get rid of it than with a janitor. The first functionality our friendly janitor has is to list out all DNS Zone delegations under `oslo.systems` that don't appear to be in use.

It comes with an accompanying Github Action that posts its result to #kjøremiljø-aktiviteter

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,9 @@ WORKDIR /build
 RUN CGO_ENABLED=0 GOOS=linux go build -a -o janitor cmd/janitor/*.go
 
 FROM alpine:3.10.8
+RUN apk add --no-cache bash
 COPY .github/actions/janitor/entrypoint.sh /entrypoint.sh
-COPY --from=builder /build/janitor /
+COPY --from=builder /build/janitor /janitor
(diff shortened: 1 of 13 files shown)


deifyed (approved):

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.16.3-alpine3.12 as builder
```
yngvark, line 1: Can we specify somewhere that this Dockerfile is janitor-only? Or call it Dockerfile-janitor? Or put it in a janitor directory?

Just something to avoid someone confusing this be an okctl-Dockerimage somehow.

yngvark (approved):
Awesome feature!

Dockerfile
paulbes: I will add a comment, github action requires it be called Dockerfile, and Docker needs the repo as context

PR #482, opened 2021-04-26
[KM240] Implement delete as a reconciler

paulbes:
Implement delete as a reconciler

## Description
- Ensure that we delete everything we store in storm
- Storing in storm is now an upsert
- Introduce managed tag on resources
- Rewrite reconcile to handle delete
- Add resource quota check for fargate pods
- Move resource quota check into reconciler
- Move cleanup into reconciler
- Store all []byte as string in storm

## Types of changes

## Checklist:


pkg/config/constant/api.go
```diff
@@ -103,6 +103,9 @@ const (
 	// DefaultRequiredIgws number of internet gateways required for cluster creation
 	DefaultRequiredIgws = 1
 
+	// DefaultRequiredFargateOnDemandPods is the minimum number of fargate pods that should be available
+	DefaultRequiredFargateOnDemandPods = 50
```
deifyed, line 107: Why 50? Just a seemingly high enough number?

pkg/controller/reconciler/postgres_group_reconciler.go
```diff
@@ -0,0 +1,32 @@
+package reconciler
+
+import (
+	clientCore "github.com/oslokommune/okctl/pkg/client/core"
+	"github.com/oslokommune/okctl/pkg/controller/resourcetree"
+)
+
+// PostgresGroupReconciler handles reconciliation for dummy nodes (e.g. the root node) and acts as a template for other
+// reconcilers
+type PostgresGroupReconciler struct{}
```
deifyed, line 10: This reconciler does nothing. Why not use the [NoopReconciler](https://github.com/oslokommune/okctl/blob/master/pkg/controller/reconciler/noop.go) and a [group type](https://github.com/oslokommune/okctl/blob/master/pkg/controller/resourcetree/tree.go#L19) to group the resources? Thats what they're for. Or are they lacking something?

pkg/controller/reconciler/sevicequota_reconciler.go
```diff
@@ -0,0 +1,74 @@
...
+
+// SetCommonMetadata knows how to store common metadata on the reconciler. This should do nothing if common metadata is
+// not needed
+func (r *ServiceQuotaReconciler) SetCommonMetadata(meta *resourcetree.CommonMetadata) {
+	r.commonMetadata = meta
+}
+
+// Reconcile knows how to create, update and delete the relevant resource
+func (r *ServiceQuotaReconciler) Reconcile(node *resourcetree.ResourceNode) (result ReconcilationResult, err error) {
+	switch node.State {
```
deifyed, line 42: A state switch seems redundant here

pkg/controller/resourcetree/tree.go
```diff
@@ -12,96 +12,62 @@ import (
...
-		return "Nameservers Delegated Test"
-	case ResourceNodeTypeUsers:
-		return "Users"
-	case ResourceNodeTypePostgres:
-		return "Postgres"
-	default:
-		return "N/A"
-	}
+// String representation of the ResourceNodeType
+func (t ResourceNodeType) String() string {
```
deifyed, line 69: Nice one :D My bad

pkg/controller/synchronize.go
```diff
@@ -40,9 +42,19 @@ func Synchronize(opts *SynchronizeOpts) error {
 	}
 
 	desiredTree.ApplyFunction(applyDeclaration(opts.ClusterDeclaration), &resourcetree.ResourceNode{})
+
+	if opts.DeleteAll {
```
deifyed, line 46: I'd suggest creating a `DeleteTree` function as an alternative to `Synchronize`. A delete tree function would be alot simpler and we wouldn't pollute the synchronize with lots of repetative if's

deifyed (approved):
👍👍👍👍

pkg/client/core/state/storm/cluster_state.go
```diff
@@ -41,18 +42,38 @@ func (c *Cluster) Convert() *client.Cluster {
...
 
 func (c *clusterState) SaveCluster(cluster *client.Cluster) error {
-	return c.node.Save(NewCluster(cluster, NewMetadata()))
+	existing, err := c.getCluster(cluster.Name)
+	if err != nil && !errors.Is(err, stormpkg.ErrNotFound) {
+		return err
+	}
+
+	if errors.Is(err, stormpkg.ErrNotFound) {
+		return c.node.Save(NewCluster(cluster, NewMetadata()))
```
yngvark, line 51: Is there a reason NewCluster is exported? (And the other New* functions in this package?)

pkg/client/core/state/storm/argocd_state.go
```diff
@@ -45,17 +46,37 @@ func (a *ArgoCD) Convert() *client.ArgoCD {
 }
 
 func (a *argoCDState) SaveArgoCD(cd *client.ArgoCD) error {
-	return a.node.Save(NewArgoCD(cd, NewMetadata()))
+	existing, err := a.getArgoCD()
```
yngvark, line 49: Would it not be possible to create a single implementation for the repeated logic used in all the x_state.go files? Or will the interface{} start spreading too much perhaps?

pkg/clusterconfig/testdata/aws-alb-ingres-controller.golden
```diff
@@ -15,5 +15,7 @@ metadata:
   name: test
   region: eu-west-1
   tags:
+    alpha.okctl.io/cluster-name: test
+    alpha.okctl.io/managed: "true"
```
yngvark, line 19: Love these new tags!

yngvark (approved):
I can't wait to try this!

pkg/config/constant/api.go
ivaruf: We can easily use 20->30 pods before we add custom applications, 50 is fairly conservative.

deifyed: How do we(or the user) increase this if the user hits 50?

paulbes: Perhaps the name is a little misleading, read this as a minimum of 50 pods should be available, the AWS default is 1000 fargate ondemand pods (calculated across ecs and eks)

A user of the AWS account would have to request a service quota increase

pkg/controller/reconciler/postgres_group_reconciler.go
paulbes: I noticed them, but did not know how to use them properly :D

pkg/client/core/state/storm/argocd_state.go
paulbes: Yes, I think this should be possible, I am simply erring on the side of caution for now, in general I have found that generalising too early can cause more pain than it is worth

pkg/client/core/state/storm/cluster_state.go
paulbes: They could probably be package scoped, but there isn't any real reason to conceal them, they aren't that internal. Also, I was thinking I might use them for the migration stuff, TBD

pkg/controller/reconciler/sevicequota_reconciler.go
paulbes: Just wanted to enumerate the scenarios the reconciler should consider, and I like switch statements more than if 🙈

pkg/controller/synchronize.go
paulbes: I would like to hear more! I am totally up for a rewrite of this stuff

cmd/okctl/delete.go
```diff
@@ -91,137 +89,83 @@ including VPC, this is a highly destructive operation.`,
...
-			err = services.Monitoring.DeleteKubePromStack(o.Ctx, client.DeleteKubePromStackOpts{
-				ID:     id,
-				Domain: domain.Domain,
-			})
-			if err != nil {
-				return formatErr(err)
+			id := api.ID{
+				Region:       opts.Region,
+				AWSAccountID: opts.AWSAccountID,
+				ClusterName:  opts.ClusterName,
```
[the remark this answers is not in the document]
paulbes: Sure, could do that 👍🏽

pkg/controller/reconciler/postgres_group_reconciler.go
deifyed: Gahhh. My bad. Should be exactly the same as you used this

PR #485, opened 2021-04-26
[KM169] Implement delete NS record delegation

paulbes:
Implement delete NS record delegation

## Description
- Add tests for ns delegation, both delete and create
- Add ability to remove ns delegation
- Add ability to automatically merge domains under auto.oslo.systems

We also introduce an experimental automatic merge:

```yaml
          experimental:
            automatizeZoneDelegation: true
```

If this section is added to the `cluster.yaml` we will add the label `automerge` to the PR. By adding a github action that looks for this label we can selectively automerge certain requests.

## Types of changes

## Checklist:


pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -290,6 +295,14 @@ func (c ClusterDatabasesPostgres) Validate() error {
 	)
 }
 
+// ClusterExperimental contains experimental fields
+type ClusterExperimental struct {
```
deifyed, line 299: I like

pkg/controller/reconciler/nameserverdelegator_reconciler.go
```diff
@@ -36,31 +38,46 @@ func (z *nameserverDelegationReconciler) SetStateHandlers(handlers *clientCore.S
...
-		if err != nil {
-			return result, fmt.Errorf("getting primary hosted zone: %w", err)
+	hz, err := z.stateHandlers.Domain.GetPrimaryHostedZone()
+	if err != nil {
+		return result, fmt.Errorf("getting primary hosted zone: %w", err)
+	}
+
+	var labels []string
+
+	if z.commonMetadata.Declaration.Experimental != nil {
```
deifyed, line 48: If we make the Experimental struct a non pointer, we wont have to check for nil. Is there any advantage of using a pointer? (except for microscopic memory usage reduction)

deifyed (approved):

ivaruf:
Won't this enable anyone to add any *.oslo.systems NS records to their hosted zone and potentially also overwrite existing records? If so this seems a bit dangerous. (will read the code now :D )

ivaruf:
> Won't this enable anyone to add any *.oslo.systems NS records to their hosted zone and potentially also overwrite existing records? If so this seems a bit dangerous. (will read the code now :D )

Never mind, seems like we make PRs for every case, and this is just a better way to handle lifecycle of hosted zones generated by okctl (I would like to get that confirmed though :+1: )

ivaruf (approved):

yngvark (approved):
Grreat!

pkg/controller/reconciler/nameserverdelegator_reconciler.go
paulbes: Since we are using `omitempty`, we need to use a pointer, otherwise it will create an empty struct

PR #492, opened 2021-04-29
✅ Add default dnsPolicy to apply application podspec

deifyed:
## Description

This seems to be required to request resources from a pod to outside of the cluster.

See https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

At a bare minimum, it is necessary to reach RDS. 

## Motivation and Context

Connecting to RDS is not possible from a pod without a dnsPolicy set.

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.56.md
+++ b/docs/release_notes/0.0.56.md
@@ -1,0 +1,6 @@
+# Release 0.0.56
+
+## Features
+
+## Changes
+✅ Add default dns policy to `apply application`
--- a/pkg/client/core/testdata/deployment.yaml.golden
+++ b/pkg/client/core/testdata/deployment.yaml.golden
@@ -19,6 +19,7 @@ spec:
         volumeMounts:
         - mountPath: /path/to/mount/volume
           name: my-app-pathtomountvolume
+      dnsPolicy: Default
       volumes:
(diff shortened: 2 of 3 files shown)


ivaruf (approved):
So nice to see a PR with 3 changed files! :D

yngvark:
Ideally, I would like to need as few "special" fields on pod.yaml as possible. Have you checked if it's possible to get the same result with using the default dns policy, which seems to "ClusterFirst"? Ref From https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy -> "Default" is not the default DNS policy. If dnsPolicy is not explicitly specified, then "ClusterFirst" is used.

But that's ideally, if we must have this, or if requires much work with the ClusterFirst, then I guess we need it.

yngvark (approved):

yngvark:
> This seems to be required to request resources from a pod to outside of the cluster.

Were you not able to reach anything on the internet?

deifyed:
@yngvark `ping google.com` failed from inside the pod. I'm no guru with networking, but to me that indicates no connection.

(I also tried the IP directly to be sure, which also failed)

PR #495, opened 2021-04-29
[KM145] Stage kubeconfig correctly in venv and show

paulbes:
Stage kubeconfig correctly in venv and show

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/venv.go
+++ b/cmd/okctl/venv.go
@@ -6,9 +6,12 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path"
 	"strings"
 
+	"github.com/oslokommune/okctl/pkg/config/constant"
 	"github.com/oslokommune/okctl/pkg/config/state"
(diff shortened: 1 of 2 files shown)


cmd/okctl/venv.go
```diff
@@ -56,6 +59,35 @@ func buildVenvCommand(o *okctl.Okctl) *cobra.Command {
 			return nil
 		},
 		RunE: func(_ *cobra.Command, args []string) error {
+			handlers := o.StateHandlers(o.StateNodes())
```
yngvark, line 62: I compared, and it seems venv.go and show.go contains 100% the same code for this kubeconfig part. Is there any issue with de-duplicating this logic?

yngvark (approved):
Good fix!

cmd/okctl/venv.go
paulbes: We should probably move into a common function, but I think we are ok with that duplication for now

PR #499, opened 2021-04-29
[KM247] Fix the cleanup reconcilers

paulbes:
The cleanup reconcilers weren't being applied correctly, due to my current understanding of the reconcilers I found it easiest to apply the desired state after the differences were calculated.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/controller/resourcetree/tree.go
+++ b/pkg/controller/resourcetree/tree.go
@@ -137,9 +137,20 @@ func (node *ResourceNode) treeViewGenerator(writer io.Writer, tabs int) {
 		result += "\t"
 	}
 
-	result += "- " + node.Type.String() + "\n"
+	result += "- " + node.Type.String()
 
 	if node.State == ResourceNodeStatePresent {
+		result += " (add)" + "\n"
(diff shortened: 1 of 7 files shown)


yngvark (approved):
Can we mention in the description of this PR if this is a refactor and why, if that is the case? It can be as simple as "Just refactoring, I want to put stuff where they belong". Makes the PR easier to understand, and it's good when looking back to it in the future.

pkg/controller/resourcetree/tree.go
```diff
@@ -137,9 +137,20 @@ func (node *ResourceNode) treeViewGenerator(writer io.Writer, tabs int) {
 		result += "\t"
 	}
 
-	result += "- " + node.Type.String() + "\n"
+	result += "- " + node.Type.String()
 
 	if node.State == ResourceNodeStatePresent {
+		result += " (add)" + "\n"
```
deifyed, line 143: 😻

deifyed (changes requested):
Hold on, making a change

deifyed (approved):

PR #515, opened 2021-05-06
[NOTASK] Slightly different approach to delete ordering

paulbes:
Slightly different approach to delete ordering

## Types of changes

## Checklist:


deifyed (commented):
I'll let others decide.

I see no advantage of doing it like this. 

The old way creates a dependency tree and operates on the tree.
This way creates a dependency tree, converts it to a list and operates on the list.

paulbes:
Perhaps it is not clear from the test output, but previously we processed a node if it had no children, even if it wasn't the deepest node. This improves a little bit on that situation, but I wouldn't call this optimal either. It is meant as a quick fix, as it ensures we remove things in a slightly better order

deifyed (approved):

PR #541, opened 2021-05-19
KM275 🐛 Allow max 5 connections for pgbouncer

yngvark:
## Description

Sometimes, pgbouncer sets up a lot of connections (up to 100) to the database (because it is in fact a connection pooler). It could potentially starve the application, depending on maximum number of connections allowed by RDS. We saw this happen for a team when using okctl forward postgres a lot.

Not sure exactly why pgbouncer establishes so many connections, but regardless, pgbbouncer should never need this anyway.

## How Has This Been Tested?

I ran okctl forward postgres in my own cluster. It works fine. Also, the pgbouncer pod's /etc/pgbouncer/pcbouncer.ini contains `max_client_conn = 5`, so it should work.

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.60.md
+++ b/docs/release_notes/0.0.60.md
@@ -3,6 +3,7 @@
 ## Features
 
 ## Bugfixes
+* KM275: Allow max 5 database connections for pgbouncer (used by okctl forward postgres) (#541)
 
 ## Changes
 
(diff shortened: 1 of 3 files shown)


deifyed (approved):

paulbes (approved):
Makes perfect sense, nice!

ivaruf (commented):
Looks good, pgbouncer does not need 100 connections

PR #573, opened 2021-06-01
Fix 👌: Use sh -c to ensure that the piped file has access to write in…

ivaruf:
One line, read the commend :D

### DIFF
--- a/cmd/okctl/completion.go
+++ b/cmd/okctl/completion.go
@@ -18,7 +18,7 @@ Bash:
 
 	# To load completions for each session, execute once:
 	# Linux:
-	$ okctl completion bash > /etc/bash_completion.d/okctl
+	$ sudo sh -c 'okctl completion bash > /etc/bash_completion.d/okctl'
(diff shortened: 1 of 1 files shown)


cmd/okctl/completion.go
```diff
@@ -18,7 +18,7 @@ Bash:
 
 	# To load completions for each session, execute once:
 	# Linux:
-	$ okctl completion bash > /etc/bash_completion.d/okctl
+	$ sudo sh -c 'okctl completion bash > /etc/bash_completion.d/okctl
```
yngvark, line 21: Add `'` in the end

yngvark (changes requested):

cmd/okctl/completion.go
ivaruf: ... damn it, I had one job

ivaruf: Fixed

yngvark (approved):
:tada:

PR #606, opened 2021-06-21
KM315 🐛 Cast to PostgresReconcilerState pointer

yngvark:
## Description

* Cast to `*clusterrec.PostgresReconcilerState` instead of `clusterrec.PostgresReconcilerState`
* I also removed the cast check because:
Today I learned that the actual error message when a cast fails is useful for debugging for the developer. With the check, the output is:

```
could not cast to database state
```

but without the check, we get the actual error

```
panic: interface conversion: interface {} is *reconciliation.PostgresReconcilerState, not reconciliation.PostgresReconcilerState
```

It was quite hard to spot the missing `*` in the IntelliJ debugger, so this error helped. UX is another thing, if we want a nice error to the user, then we should probably print something nicer.

Not adding tests because [KM305](https://trello.com/c/ws2w0db9/305-clean-up-tree-structure) will rewrite the whole reconciliation soon.

## How Has This Been Tested?

* Ran okctl apply cluster on my own cluster, this change fixed the issue
* Ran existing unit tests

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.63.md
+++ b/docs/release_notes/0.0.63.md
@@ -10,6 +10,8 @@ KM302 🐛 Fix broken validation for vpc.highAvailability setting (#580)
 
 KM309 🐛 Fix broken access key authentication type for apply cluster (#593)
 
+KM315 🐛 Fix reconciliation of Postgres breaking (#606)
+
 ## Changes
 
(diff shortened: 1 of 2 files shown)


deifyed (approved):
nice catch!

PR #608, opened 2021-06-22
KM305 👌 Improve reconciliation functionality

deifyed:
## Description

This PR:
- Replaces old tree based reconciliation scheduling with a new queue based scheduling
- Simplifies Reconciler interface by 
    - Cleaning up Metadata
    - Moving Metadata to Reconcile() parameter
- Simplifies spinner (though removes duration)
- Adds delete user functionality
- Cleans up some misuse of the Reconciler concept (places where the reconciler should handle multiple resources as in Users and Postgres)
- Adds the concept of Direct State. This is state retrieved directly from the source. For cluster resources it means either using `kubectl` or `helm` to determine existence. For CFN resources it means calling the AWS API

## Motivation and Context

The complexity of the current implementation is inducing a number of misimplementations and smaller bugs

[KM305](https://trello.com/c/ws2w0db9)
[KM292](https://trello.com/c/ZysQ4IMF)
[KM300](https://trello.com/c/JDFpiPBX)
[KM244](https://trello.com/c/8oPca592)

## Types of changes

## Checklist:


pkg/api/core/cloudprovider/aws/identitymanager_aws.go
```diff
@@ -32,6 +32,17 @@ func (s *identityManagerCloudProvider) DeleteIdentityPoolClient(opts api.DeleteI
 	return nil
 }
 
+func (s *identityManagerCloudProvider) DeleteIdentityPoolUser(opts api.DeleteIdentityPoolUserOpts) error {
```
ivaruf, line 35: Nice, now we have delete user? :)

pkg/controller/cluster/reconciliation/argocd_reconciler_test.go
```diff
@@ -0,0 +1,188 @@
+package reconciliation
```
yngvark, line 1: Unrelated to this file: The test coverage is quite good, although when I was able to run all reconciler tests with coverage, I see some uncovered code that I think would be useful to test. The most common untested case seems to be the Noop case, which would be the result of two test cases: Do nothing when user doesn't want the component, and it doesn't exist - and vice versa, do nothing when the user wants the component and already has it.

Other than that, most of the errors are not tested, but unless there is some specific error handling other than just returning the error, not testing those is fine.

yngvark (changes requested):
For the amount of code changed, the cleanliness and apprehensibility of this PR is remarkable, which also is a good indicator that we're refactoring in a good direction. Nice job!

pkg/controller/cluster/reconciliation/cleanup_alb_reconciler.go
```diff
@@ -1,67 +0,0 @@
...
-		return result, nil
-	case dependencytree.NodeStateAbsent:
-		vpc, err := state.Vpc.GetVpc(cfn.NewStackNamer().Vpc(z.commonMetadata.ClusterID.ClusterName))
-		if err != nil {
-			return result, fmt.Errorf("getting vpc: %w", err)
-		}
-
-		clean := cleaner.New(z.provider)
-
-		err = clean.DeleteDanglingALBs(vpc.VpcID)
```
ivaruf, line 48: We should check if we still need this when we get delete to work as intended. Hopefully there will be no need to delete any dangling ALBs or security groups. (Out of scope for this PR though)

pkg/okctl/okctl.go
```diff
@@ -140,23 +142,32 @@ func (o *Okctl) StateNodes() *clientCore.StateNodes {
...
+		Vpc:                       storm.NewVpcState(nodes.Vpc),
+		Parameter:                 storm.NewParameterState(nodes.Parameter),
+		Domain:                    storm.NewDomainState(nodes.Domain),
+		ExternalDNS:               storm.NewExternalDNSState(nodes.ExternalDNS),
+		Cluster:                   storm.NewClusterState(nodes.Cluster),
+		Component:                 storm.NewComponentState(nodes.Component),
+		Monitoring:                storm.NewMonitoringState(nodes.Monitoring),
+		ArgoCD:                    storm.NewArgoCDState(nodes.ArgoCD),
+		ContainerRepository:       storm.NewContainerRepositoryState(nodes.ContainerRepository),
+		Loki:                      direct.NewLokiState(o.Declaration.Metadata, helmClient),
```
ivaruf, line 164: Very nice that we have state for Loki, Promtail etc now

ivaruf (approved):
Phenomenal work, all of this was overdue for a proper refactoring 😄

pkg/api/core/cloudprovider/aws/identitymanager_aws.go
deifyed: Heck yes

pkg/controller/cluster/reconciliation/cleanup_alb_reconciler.go
deifyed: Yeah. We agreed to postpone it so that the PR would be ready for review before the summer though. This reconciler is a thorn in my side, so I'd love to remove it and fix things properly at some point

pkg/okctl/okctl.go
deifyed: 💪

pkg/controller/application/reconciliation/application_reconciler.go
```diff
@@ -1,77 +1,139 @@
 package reconciliation
```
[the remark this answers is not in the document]
deifyed: Conclusion from whereby chat:
* application_reconciler and container_repository_reconciler: I'll refactor apply application pattern to better match the apply cluster pattern after the summer. I'll postpone creating tests for these until then
* nameserverdelegator tests; I'm pondering a different design here, awaiting conclusion of that

yngvark (approved):

PR #684, opened 2021-09-10
KM-379: Change scaffold of volumes to use ReadWriteOnce instead of Many

ivaruf:
ReadWriteMany is for multiple nodes to be connected to the same PVC
It will not work in this case
https://kubernetes.io/docs/concepts/storage/persistent-volumes/

### DIFF
--- a/pkg/client/core/testdata/volumes.yaml.golden
+++ b/pkg/client/core/testdata/volumes.yaml.golden
@@ -5,7 +5,7 @@ metadata:
   namespace: okctl
 spec:
   accessModes:
-  - ReadWriteMany
+  - ReadWriteOnce
   resources:
     requests:
       storage: 24Gi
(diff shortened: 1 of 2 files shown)


yngvark:
"It will not work in this case" - I'm guessing this refer to
"Note that an EBS volume can only be used by a single pod at the same time. Thus, the access mode of your PVC can only be ReadWriteOnce."

https://docs.giantswarm.io/advanced/storage/ebs-csi/

Though I wasn't able to find this limitation in the AWS official docs (by quick googling).

yngvark (approved):
"It will not work in this case" - I'm guessing this refer to
"Note that an EBS volume can only be used by a single pod at the same time. Thus, the access mode of your PVC can only be ReadWriteOnce.

?

https://docs.giantswarm.io/advanced/storage/ebs-csi/

Though I wasn't able to find this limitation in the AWS official docs (by quick googling).

ivaruf:
Mostly "Will not work in this case" was : I tested it , it failed. Could have included the error message, but I didn't :sweat_smile:

PR #694, opened 2021-09-28
KM287 ✅ Adjust upgrade flags

yngvark:
Update upgrade binary flags semantics.

## Description

Dependent on: https://github.com/oslokommune/okctl-upgrade/pull/3

* Added support for flags described in https://github.com/oslokommune/okctl-upgrade/pull/3
  * The essence of this PR changes is in okctlupgrade.go. Everything else is just support and refactoring.
* Refactoring

(61 out of 82 (atm) changed files is testdata, so only 21 code files changed.)

## How Has This Been Tested?

* New unit tests.
* Running upgrade binaries asking for user input is somewhat hard to test, so I have skipped that for now.
* Manual tests. 

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
  - Since feature is still hidden, I haven't edited these. When feature is un-hidden, I'll make a release note.


pkg/upgrade/upgrade.go
```diff
@@ -217,6 +176,51 @@ func (u Upgrader) dryRunBinaries(upgradeBinaries []okctlUpgradeBinary, binaryPro
...
+			return fmt.Errorf("marking upgrades as run: %w", err)
+		}
+
+		// Update cluster version
+		// Note that we don't save the hotfix version, only the semver version. This is because we use the cluster
+		// versioner later to validate that the current okctl version against the cluster version. The okctl version
+		// will always be a semver version without any hotfix in it, so it wouldn't be possible to store a hotfix
+		// version here.
+		//
+		// Also, a hotfix version is supposed to just fix any errors being made in an upgrade binary, so
```
deifyed, line 212: Nice to know. This means that we can only create one upgrade per okctl release. This should be communicated somewhere else as well. Maybe a short RFC

deifyed (approved):

pkg/upgrade/upgrade.go
yngvark: Agh, that was not the intention. We ideally want to be able to have as many upgrades as we want for a release, for nice separation.

I'll have to look into it.

yngvark: As discussed in Slack, we do support multiple upgrades for one okctl version, because we support hotfixes.

PR #697, opened 2021-10-04
argocd application controller: increase K8s resources

olovholm:
## Description
Extends the resource request and limitations for argocd application controller

## Motivation and Context
We experiences that the argocd application controller used more resources than the allocated out-of-the-box defined values which caused several restarts from OOM challenges for the container. We solved this problem, in our test cluster, by doubling the resources requested. In this PR we also double the limit for memory and increase CPU from 500 to 800.

## How Has This Been Tested?
By doubling the requested resources in our test environment we were able to avoid the problem, and our application controller container has been running without any resource challenges for the last three days (from 01.10).

## Types of changes

## Checklist:


yngvark (approved):
Good catch, thanks for the contribution!

PR #702, opened 2021-10-06
🐛 Fix breaking test after k3d-proxy upgrade

yngvark:
After release of k3d 5.0.0, k3d-proxy docker image fails.

Not sure whether it's a bug or not, but we should anyway not
use latest versions.

## Types of changes

## Checklist:


pkg/integration/integration.go
```diff
@@ -23,6 +23,7 @@ import (
 	k3dCluster "github.com/rancher/k3d/v3/pkg/cluster"
 	"github.com/rancher/k3d/v3/pkg/runtimes"
 	k3d "github.com/rancher/k3d/v3/pkg/types"
+	"github.com/rancher/k3d/v3/version"
```
deifyed, line 26: Nitpick: Maybe use k3dVersion to follow the (for all I know, only this file) naming pattern

deifyed (approved):
Nice catch and quick response. All the claps

PR #703, opened 2021-10-06
KM406 👌 Use semver for local versioning

yngvark:
version.GetVersionInfo should return a version on semver format, for instance "0.0.32", not "dev".

## Description

Upgrade functionality is dependent on using a semver version, so if we use "dev" when doing okctl operations locally (before releasing), stuff will break. Right now, we have a validation that checks that binary version is not less then cluster version. So if binary version is "dev", this comparison breaks.

This PR solves this, by checking that if version is dev, replace it with latest github release version, or "0.0.10" if github somehow doesn't respond.

## Motivation and Context

See above.

Run `okctl version`, where okctl is a local release, and verify that the version is not `dev`, but the same version as the latest github release of this repository.

## Types of changes

## Checklist:


yngvark:
Messed commits up by forking a discarded branch. See #704 instead.

PR #733, opened 2021-10-27
POC: Ask k8s if namespace exists before creating manifest file

ivaruf:
This is a proof of concept of how we could solve this bug. We obviously will not merge this since the code calls 'kubctl' directly in scaffold/api.

Basically we just skip writing the namespace.yaml manifest file if namespace for app already exists. 

Proposed POC works in the code when called from venv (for some reason the same code will not be called when not called from venv? I did not investigate this and it's beside the point).

The main problem with this approach is the tight coupling between application and namespace. If for example you delete the first application that has the original (namespace.yaml) manifest, then the other application that use this namespace will cease to function.

Please discuss :)

### DIFF
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -2,7 +2,10 @@
 package scaffold
 
 import (
+	"bufio"
 	"fmt"
+	"os/exec"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


deifyed:
`delete application` could solve this by deleting all resources with `app=<app name` label and see if theres anything left; if not then delete namespace, else leave as is

PR #748, opened 2021-11-04
Document how to upgrade application

yngvark:
First sketch for how to upgrade an application.

## Description

Before getting QA from UX, I would like comments if we want to document it this way.

## Motivation and Context

The users' k8s manifests sometimes needs to be upgraded, for instance when Ingress goes from v1beta to v1. Instead of making complex upgrade that attempts to upgade users k8s manifests, which they can have rearranged in ways unknown to us, we in this guide inform the user how to do the upgrade themselves.

If we attempt to create automatic upgrades for k8s manifests, it's never going to be perfect, it would always be best-effort, which is messy.

Doing it the way described in this guide means more work for the user, but it also requires them to have some understanding of their manifests, which we want. 

Open the netlify link and see the new document Application -> Upgrading.

## Types of changes

## Checklist:


kmoberg (approved):
Nice!

kmoberg (approved):

deifyed (approved):
I like.

PR #758, opened 2021-11-11
KM434 🐛 Add control plane SecurityGroup to security group policy

yngvark:
Adds the control plane security group to `infrastructure/application/some-app/overlays/securitygrouppolicy-patch.json`.

## Description

This PR provides a step to fixing cluster communication between services, which currently doesn't work, while using security groups.

## Motivation and Context

[KM434](https://trello.com/c/N6PTVzf4/434-fix-communication-for-pods-with-security-group)

Run `okctl -c my-cluster.yaml apply application -f my-app.yaml`. The output file `securitygrouppolicy-patch.json` should include two security groups. One of them should be the same as running

```
aws eks describe-cluster --name <cluster_name> --query cluster.resourcesVpcConfig.clusterSecurityGroupId
```

This is the same as the security group with the description

> EKS created security group applied to ENI that is attached to EKS Control Plane master nodes, as well as any managed workloads.

in EC2->Security groups.

Source: https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (See: https://github.com/oslokommune/okctl/commit/446e412a110c05f841e526045625cb794ebf044e)

### DIFF
--- a/pkg/api/cluster_api.go
+++ b/pkg/api/cluster_api.go
@@ -72,7 +72,7 @@ func (o *ClusterSecurityGroupIDGetOpts) Validate() error {
 type ClusterService interface {
 	CreateCluster(context.Context, ClusterCreateOpts) (*Cluster, error)
 	DeleteCluster(context.Context, ClusterDeleteOpts) error
-	GetClusterSecurityGroupID(context.Context, ClusterSecurityGroupIDGetOpts) (*ClusterSecurityGroupID, error)
+	GetClusterSecurityGroupID(context.Context, *ClusterSecurityGroupIDGetOpts) (*ClusterSecurityGroupID, error)
 }
 
 // ClusterRun provides an interface for running CLIs
--- a/pkg/api/cluster_api.go
+++ b/pkg/api/cluster_api.go
@@ -50,10 +50,29 @@ func (o *ClusterDeleteOpts) Validate() error {
 	)
 }
 
+// ClusterSecurityGroupIDGetOpts specifies the required inputs for getting the cluster's security group
+type ClusterSecurityGroupIDGetOpts struct {
+	ID ID
+}
+
+// ClusterSecurityGroupID contains an EKS cluster's cluster security group ID
+// See https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html
+type ClusterSecurityGroupID struct {
+	Value string
+}
+
+// Validate the delete inputs
+func (o *ClusterSecurityGroupIDGetOpts) Validate() error {
+	return validation.ValidateStruct(o,
+		validation.Field(&o.ID, validation.Required),
+	)
+}
+
 // ClusterService provides an interface for the business logic when working with clusters
 type ClusterService interface {
 	CreateCluster(context.Context, ClusterCreateOpts) (*Cluster, error)
 	DeleteCluster(context.Context, ClusterDeleteOpts) error
+	GetClusterSecurityGroupID(context.Context, ClusterSecurityGroupIDGetOpts) (*ClusterSecurityGroupID, error)
 }
 
 // ClusterRun provides an interface for running CLIs
--- a/pkg/api/cluster_api.go
+++ b/pkg/api/cluster_api.go
@@ -70,8 +70,18 @@ func (o *ClusterSecurityGroupIDGetOpts) Validate() error {
 
 // ClusterService provides an interface for the business logic when working with clusters
 type ClusterService interface {
+	ClusterCruder
+	ClusterDetailer
+}
+
+// ClusterCruder knows how to create and delete clusters
+type ClusterCruder interface {
 	CreateCluster(context.Context, ClusterCreateOpts) (*Cluster, error)
 	DeleteCluster(context.Context, ClusterDeleteOpts) error
+}
+
+// ClusterDetailer knows how to get details about a cluster
+type ClusterDetailer interface {
 	GetClusterSecurityGroupID(context.Context, *ClusterSecurityGroupIDGetOpts) (*ClusterSecurityGroupID, error)
 }
 
--- a/pkg/api/core/handler_api.go
+++ b/pkg/api/core/handler_api.go
@@ -391,6 +391,7 @@ type EndpointOption func(Endpoints) Endpoints
 
 const (
 	clusterTag             = "clusterService"
+	securityGroupIDTag     = "securityGroupID"
 	vpcTag                 = "vpc"
 	managedPoliciesTag     = "managedPolicies"
 	externalSecretsTag     = "externalSecrets"
@@ -429,7 +430,7 @@ func InstrumentEndpoints(logger *logrus.Logger) EndpointOption {
 		return Endpoints{
 			CreateCluster:                   logmd.Logging(logger, "create", clusterTag)(endpoints.CreateCluster),
 			DeleteCluster:                   logmd.Logging(logger, "delete", clusterTag)(endpoints.DeleteCluster),
-			GetClusterSecurityGroupID:       logmd.Logging(logger, "get", clusterTag, "securityGroupID")(endpoints.GetClusterSecurityGroupID),
+			GetClusterSecurityGroupID:       logmd.Logging(logger, "get", clusterTag, securityGroupIDTag)(endpoints.GetClusterSecurityGroupID),
 			CreateVpc:                       logmd.Logging(logger, "create", vpcTag)(endpoints.CreateVpc),
 			DeleteVpc:                       logmd.Logging(logger, "delete", vpcTag)(endpoints.DeleteVpc),
 			CreateExternalDNSKubeDeployment: logmd.Logging(logger, "create", kubeTag, externalDNSTag)(endpoints.CreateExternalDNSKubeDeployment),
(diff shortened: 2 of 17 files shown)


pkg/client/core/service_application_postgres_client.go
```diff
@@ -208,7 +247,12 @@ func (a *applicationPostgresService) generateSecurityGroupPolicy(ctx context.Con
 				{
 					Type:  jsonpatch.OperationTypeAdd,
 					Path:  "/spec/securityGroups/groupIds/0",
-					Value: sg.ID,
+					Value: clusterSecurityGroupID,
```
yngvark, line 250: This is the actual change of this PR

pkg/client/core/api/rest/cluster_rest.go
```diff
@@ -21,6 +24,11 @@ func (a *clusterAPI) DeleteCluster(opts api.ClusterDeleteOpts) error {
 	return a.client.DoDelete(ClusterTarget, &opts)
 }
 
+func (a *clusterAPI) GetClusterSecurityGroupID(opts api.ClusterSecurityGroupIDGetOpts) (*api.ClusterSecurityGroupID, error) {
+	id := &api.ClusterSecurityGroupID{}
+	return id, a.client.DoGet(ClusterSecurityGroupIDTarget, &opts, &id)
```
deifyed, line 29: Pointer to a pointer? I don't think thats necessary

pkg/api/cluster_api.go
```diff
@@ -50,10 +50,29 @@ func (o *ClusterDeleteOpts) Validate() error {
...
+	return validation.ValidateStruct(o,
+		validation.Field(&o.ID, validation.Required),
+	)
+}
+
 // ClusterService provides an interface for the business logic when working with clusters
 type ClusterService interface {
 	CreateCluster(context.Context, ClusterCreateOpts) (*Cluster, error)
 	DeleteCluster(context.Context, ClusterDeleteOpts) error
+	GetClusterSecurityGroupID(context.Context, ClusterSecurityGroupIDGetOpts) (*ClusterSecurityGroupID, error)
```
deifyed, line 75: Thinking back on the discussion about the problem with thingsTM having to implement all kinds of strange different things.

The recommended way of structuring interfaces in Go is to divide functionality into smaller interfaces. In this case adding an interface for cluster crud operations, ClusterCruder, and an interface f.ex ClusterDetailer for functionality like `GetClusterSecurityGroupID`. ClusterService would then extend on both of them. That way functionality only requiring cluster crud operations could accept the ClusterCruder interface instead of the whole service, and the service could act as a ClusterCruder.

That said, I feel the right approach here would actually be to add the security group ID to the cluster struct, and implement GetCluster() functionality.

deifyed (changes requested):
By the way, the new security group ID wont get added to an app with a pre-existing postgres integration due to `client/core/service_application_postgres_client.go:136`. The reconciler checks this function to see if anything needs doing. 

That said; awesome job fixing this issue. That was a hard one :D

pkg/api/core/handler_api.go
```diff
@@ -253,6 +257,11 @@ func AttachRoutes(handlers *Handlers) http.Handler {
 			r.Method(http.MethodPost, "/", handlers.CreateCluster)
 			r.Method(http.MethodDelete, "/", handlers.DeleteCluster)
 		})
+
+		r.Route("/clustersecuritygroupid", func(r chi.Router) {
```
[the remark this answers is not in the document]
yngvark: That was my inital thought. But I was unsure how RESTful that would be, as I haven't seen such an endpoint before. The normal REST way would be:
/clusters/{clusterId}/securitygroupid.

... which we can't do, since don't have a clusterId, only the ID struct, which consists of three fields.

Since we pass a struct instead of just a clusterId string, perhaps your suggestion is okay.

pkg/api/core/service_cluster_api.go
```diff
@@ -40,9 +44,27 @@ func (s *clusterService) DeleteCluster(_ context.Context, opts api.ClusterDelete
...
+// GetClusterSecurityGroupID returns the EKS cluster's security group ID.
+// See https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html
+func (s *clusterService) GetClusterSecurityGroupID(
+	ctx context.Context, opts api.ClusterSecurityGroupIDGetOpts,
+) (*api.ClusterSecurityGroupID, error) {
+	cluster, err := s.cloudProvider.EKS().DescribeClusterWithContext(ctx, &eks.DescribeClusterInput{
+		Name: &opts.ID.ClusterName,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("describing cluster: %w", err)
```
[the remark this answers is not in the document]
yngvark: True, completely didn't think of that.

pkg/client/core/api/rest/cluster_rest.go
yngvark: * Insert xzibit here *

Not sure why it worked though. Fixed anyway.

pkg/api/core/handler_api.go
```diff
@@ -197,6 +200,7 @@ func MakeHandlers(responseType EncodeResponseType, endpoints Endpoints) *Handler
 	return &Handlers{
 		CreateCluster:                   newServer(endpoints.CreateCluster, decodeClusterCreateRequest),
 		DeleteCluster:                   newServer(endpoints.DeleteCluster, decodeClusterDeleteRequest),
+		GetClusterSecurityGroupID:       newServer(endpoints.GetClusterSecurityGroupID, decodeGetClusterSecurityGroupIDRequest),
```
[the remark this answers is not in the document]
yngvark: Nice, didn't see that one. Fixed.

pkg/api/cluster_api.go
yngvark: > That said, I feel the right approach here would actually be to add the security group ID to the cluster struct, and implement GetCluster() functionality.

Yeah, I wouldn't mind having that function. I did took a look at how to implement GetCluster, though. That function would return an `api.Cluster` struct, which contains lots of data, which we currently have in state. We're trying not to use state, so a GetCluster should probably retrieve all the info from AWS. It would need that actually, since the state doesn't contain the cluster security group ID, needed by this PR. (Or we could do both, merging some from state and the cluster security group ID from AWS, but that's a bit messy.)

I don't have the full picture, but I think this needs to be done:
* Implement GetCluster to get all info in the `api.Cluster` struct from AWS API calls. (`aws eks describe-cluster` gives a some of the data, but not all.)
* Replace calls to cluster_client.go -> ClusterState.{GetCluster,HasCluster,...), about 15 calls. Perhaps this could be done later, but I don't want us to have multiple sources of truth.
* Consider caching GetCluster results if possible for short amount of time, it might be a bit resource intensive to fetch all that info for every call.

So that's a bit of work. For something as central as GetCluster I see the point of implementing it, though. So if we want to prioritize it, I can give it a go. However, I think it should be done in a separate PR for cleaner separation, and since this solution is working.

> The recommended way of structuring interfaces in Go is to divide functionality into smaller interfaces.

Yeah, will do.

pkg/api/core/handler_api.go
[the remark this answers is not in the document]
deifyed: Yeah. It makes sense to me atleast to stick to the RESTful style, but with parameters in the structs instead. At least that will give us a consistent style to follow

pkg/api/cluster_api.go
deifyed: > Replace calls to cluster_client.go -> ClusterState.{GetCluster,HasCluster,...), about 15 calls. Perhaps this could be done later, but I don't want us to have multiple sources of truth.

I'd definitely save this for later. Totally agree with not wanting to have multiple ways of doing the same thing, but I'd rather eat that elephant one bite at a time. Each function we implement that directly acquires information from the API saves us strange bugs down the line. IMO

> Consider caching GetCluster results if possible for short amount of time, it might be a bit resource intensive to fetch all that info for every call.

We probably need caching at some point yes, but for now we're ~5~ teams with what, 10 calls to the API each? My suggestion of a cache implementation would be to implement a `api.CachedClusterClient` which took the original cluster client and memoized its functions

> So that's a bit of work. For something as central as GetCluster I see the point of implementing it, though. So if we want to prioritize it, I can give it a go. However, I think it should be done in a separate PR for cleaner separation, and since this solution is working.

I'd say don't prioritize it unless you want to experiment. The thoughts I've communicated here are my default approach to issues like this, and therefore doesn't cost as much time. My suggestion is either to do it this way the next time we touch similar functionality, or if you disagree, discuss it with me. 

> However, I think it should be done in a separate PR for cleaner separation, and since this solution is working.

Totally agree.

deifyed (approved):

pkg/api/core/handler_api.go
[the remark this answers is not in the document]
yngvark: /clustersecuritygroupid doesn't scale so well, i.e. if all sorts of attributes are put on root level, it's going to be messy. So I'll go for the non-RESTful /clusters/securitygroupid.

PR #786, opened 2021-11-25
KM445 👌 Add validations for okctl vs cluster version

yngvark:
Makes sure okctl binary version is equal or greater than cluster version, depending which command is being run.

### Added validations

I have added these validations for some commands.

I did not add validations for these commands:
* scaffold application -c somecluster.yaml: because it doesn't do `o.Initialize()`, meaning state handlers hasn't been initialized so we can get cluster state. We could add it, but then the user has to login to the cluster just to do a scaffold.
* version: We don't want to require cluster login just to get version
* completion: We don't want to require cluster login just to get completion
* scaffold cluster: N/A, version doesn't exist

### Removed brew support
This PR also removes brew support for okctl. The rationale is that since okctl isn't backwards compatible* with clusters of older versions, the okctl binary must either be equal or greater than the cluster version at all times (depending which command is being run). When the user decides to upgrade the cluster, and only then, should the user get a new version of okctl.

By using brew, as soon as the user runs `brew update && brew upgrade` (and there is a new version of okctl available), many commands will stop working, because of the new validations in this PR. This is not the user experience we want, we want the user to decide when to upgrade the cluster (i.e. running `okctl upgrade`).

Also, we want the user to be able to install a specific version of okctl. This is needed for instance when a user wants to use a cluster set up by a previous version of okctl. It is possible to pin versions in brew (`brew pin`), meaning it doesn't get upgraded automatically. However, there [isn't a simple way](https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula) of installing specific versions of a package with brew. `brew extract` is reported to work, but requires enabling "developer mode", which makes brew behave differently, and thus has unintented side effects.

There could be other uses cases as well, and not being able to easily switch versions isn't great. We could instead choose to create `ovm` - okctl version manager, similar to `nvm`, so the user just types `ovm install 0.0.70`. Or take it one step further and somehow implement `okctl install 0.0.70`.

\* We have decided for now to not attempt to make okctl backwards compatible for `okctl apply` commands, because this adds lots of complexity: We would have to introduce git flows for managing major, minor and patch versions, and for every new commit, decide where it should go. This process is also more prone to errors.

## Motivation and Context

[KM445](https://trello.com/c/3ViH3EKz/445-upgrade-new-validation-okctl-binary-version-must-be-equal-to-cluster-version)

1. Set cluster version: Edit cluster version in state.db with boltbrowser. Set upgrade ->  ClusterVersion -> Value to something.
1. Set okctl version: Either build (using ldflags to set version) okctl with a version less than your cluster version, or set release.go -> Version to some semantic version.
1. Then go ahead and run the commands in this PR to see that the validations work (note there are two different version validations).

To test the installation script, run

```shell
curl -fsSL https://raw.githubusercontent.com/oslokommune/okctl/KM445-okctl_version_must_equal_cluster_version/install/install.sh | /bin/bash
```

or

```shell
curl -fsSL https://raw.githubusercontent.com/oslokommune/okctl/KM445-okctl_version_must_equal_cluster_version/install/install.sh | /bin/bash -s 0.0.74
```

## Types of changes

## Checklist:

### DIFF
--- a/docs/release_notes/0.0.75.md
+++ b/docs/release_notes/0.0.75.md
@@ -14,4 +14,3 @@ KM445 👌 Add validations for okctl vs cluster version (#786).
 * **NOTE to OS X / `brew` users:** To support a proper UX for upgrades, we have removed `brew` support from okctl. To get future versions of `okctl`, see https://okctl.io/getting-started/install/. See more details in #786.
 
 ## Other
-
--- a/docs/release_notes/0.0.75.md
+++ b/docs/release_notes/0.0.75.md
@@ -10,5 +10,8 @@ KM449 🐛 Make it possible to import okctl as a library
 
 KM406 👌 Use latest okctl github release as okctl development version (#765)
 
+KM445 👌 Add validations for okctl vs cluster version (#786).
+* **NOTE to OS X / `brew` users:** To support a proper UX for upgrades, we have removed `brew` support from okctl. To get future versions of `okctl`, see https://okctl.io/getting-started/install/. See more details in #786.
+
 ## Other
 
--- a/install/install.sh
+++ b/install/install.sh
@@ -84,7 +84,7 @@ function publish_stop() {
   publish_start_stop $PHASE_KEY
 }
 
-publish_start
+# publish_start
 
 if [[ -z $1 ]]; then
   VERSION=latest
@@ -96,7 +96,7 @@ fi
 if command -v brew &> /dev/null; then
   # Check if okctl exists
   if brew list okctl &> /dev/null; then
-    publish_event brew_uninstall
+    # publish_event brew_uninstall
 
     echo Uninstalling okctl from brew
     brew uninstall okctl
@@ -112,4 +112,4 @@ else
   install_usr $VERSION
 fi
 
-publish_stop
+# publish_stop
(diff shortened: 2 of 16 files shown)


docs/release_notes/0.0.75.md
```diff
@@ -10,5 +10,8 @@ KM449 🐛 Make it possible to import okctl as a library
 
 KM406 👌 Use latest okctl github release as okctl development version (#765)
 
+KM445 👌 Add validations for okctl vs cluster version (#786).
```
kmoberg, line 13: Good brief introduction. I suggest we flag it as a breaking change as well when releasing.

kmoberg (approved):
Looks good. Tested the install procedure on macOS and it's working well! Upgrade looks good, but someone else should double-check this! Nice work! 🥳

docs/release_notes/0.0.75.md
yngvark: Yeah, I'm thinking writing a heads up in Slack as well.

install/install.sh
```diff
@@ -0,0 +1,68 @@
...
+}
+
+if [[ -z $1 ]]; then
+  VERSION=latest
+else
+  VERSION=$1
+fi
+
+# Check if brew exists
+if command -v brewlol &> /dev/null; then
```
deifyed, line 52: brewlol? :p

deifyed (approved):

install/install.sh
yngvark: yikes, local testing. Fixing.

PR #797, opened 2021-12-03
KM457 Proof-of-concept for new logging

fredriv:
New package for structured logging. 

## Description

The new logging package provides a generic interface for structured logging, as well as an implementation based on logrus. 

Currently a preview/proof-of-concept to get feedback, a lot remains e.g. configuration of logrus and logging to file.

## Motivation and Context

This will allow us to more easily add logging to the new service layer now that we are removing the client-server split, without introducing a logrus dependency everywhere. We can later implement the logging based on a different library (e.g. zap or zerolog) without changing the rest of the code.

https://trello.com/c/efJRIAuS/457-add-new-logging-solution

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/main.go
+++ b/cmd/okctl/main.go
@@ -47,6 +47,10 @@ func buildRootCommand() *cobra.Command {
 	var outputFormat string
 
 	o := okctl.New()
+	if err := o.InitLogging(); err != nil {
+		fmt.Println("Error configuring logging: %w", err)
+		os.Exit(1)
+	}
 
 	cmd := &cobra.Command{
 		Use:          "okctl",
--- a/cmd/okctl/main.go
+++ b/cmd/okctl/main.go
@@ -48,7 +48,7 @@ func buildRootCommand() *cobra.Command {
 
 	o := okctl.New()
 	if err := o.InitLogging(); err != nil {
-		fmt.Println("Error configuring logging: %w", err)
+		fmt.Fprintln(os.Stderr, "Error configuring logging:", err)
 		os.Exit(1)
 	}
 
--- a/pkg/logging/api.go
+++ b/pkg/logging/api.go
@@ -5,8 +5,13 @@ package logging
 var logger Logger //nolint: gochecknoglobals
 
 // InitLogger initialize the global logger instance
-func InitLogger(logFile string) {
-	initLogger(newLogrusLogger(logFile))
+func InitLogger(logFile string) error {
+	logger, err := newLogrusLogger(logFile)
+	if err == nil {
+		initLogger(logger)
+	}
+
+	return err
 }
 
 func initLogger(l Logger) {
--- a/pkg/logging/api.go
+++ b/pkg/logging/api.go
@@ -6,7 +6,11 @@ var logger Logger //nolint: gochecknoglobals
 
 // InitLogger initialize the global logger instance
 func InitLogger() {
-	logger = newLogrusLogger()
+	initLogger(newLogrusLogger())
+}
+
+func initLogger(l Logger) {
+	logger = l
 }
 
 // GetLogger returns a logger with the given component and activity as
--- a/pkg/logging/api.go
+++ b/pkg/logging/api.go
@@ -5,8 +5,8 @@ package logging
 var logger Logger //nolint: gochecknoglobals
 
 // InitLogger initialize the global logger instance
-func InitLogger() {
-	initLogger(newLogrusLogger())
+func InitLogger(logFile string) {
+	initLogger(newLogrusLogger(logFile))
 }
 
 func initLogger(l Logger) {
--- a/pkg/logging/api.go
+++ b/pkg/logging/api.go
@@ -1,0 +1,16 @@
+// Package logging API for structured logging
+package logging
+
+// The global logger instance
+var logger Logger //nolint: gochecknoglobals
+
+// InitLogger initialize the global logger instance
+func InitLogger() {
+	logger = newLogrusLogger()
+}
+
+// GetLogger returns a logger with the given component and activity as
+// context fields on all log entries
+func GetLogger(component string, activity string) Logger {
+	return logger.WithField("component", component).WithField("activity", activity)
+}
--- a/pkg/logging/logrus.go
+++ b/pkg/logging/logrus.go
@@ -1,0 +1,62 @@
+package logging
+
+import (
+	"os"
+
+	"github.com/oslokommune/okctl/pkg/context"
+	"github.com/sirupsen/logrus"
+)
+
+type logrusLogger struct {
+	logger *logrus.Logger
+	fields logrus.Fields
+}
+
+func (l logrusLogger) Debug(msg string) {
+	l.logger.WithFields(l.fields).Debug(msg)
+}
+
+func (l logrusLogger) Info(msg string) {
+	l.logger.WithFields(l.fields).Info(msg)
+}
+
+func (l logrusLogger) Warn(msg string) {
+	l.logger.WithFields(l.fields).Warn(msg)
+}
+
+func (l logrusLogger) Error(msg string) {
+	l.logger.WithFields(l.fields).Error(msg)
+}
+
+func (l logrusLogger) WithField(key string, value interface{}) Logger {
+	newFields := make(logrus.Fields)
+	for k, v := range l.fields {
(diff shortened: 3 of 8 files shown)


pkg/logging/types.go
```diff
@@ -0,0 +1,10 @@
+package logging
+
+type Logger interface {
+	Debug(msg string)
+	Info(msg string)
+	Warn(msg string)
+	Error(msg string)
```
deifyed, line 7:
```suggestion
	Debug(args ...interface{})
	Info(args ...interface{})
	Warn(args ...interface{})
	Error(args ...interface{})
```

We should probably allow for other types besides string. If I remember correctly, the loggers follows the signature of fmt's `Print()` function. As in `Print(a ...interface{}`.

In the logrus wrapper, it would look like this:
image

deifyed (changes requested):
A comment about the function signatures, other than that it looks good

pkg/logging/types.go
fredriv: Yes, but if we switch to e.g. `Zap` logging library it only accepts a single message (string) + a set of key-value pairs for structured logging, which we support with `WithField`.

What do you want multiple arguments for here? To avoid having to format your message yourself?

fredriv: You can always use `fmt.Sprint` or `Sprintf` for that too.

fredriv: Never mind, I see that `Zap` also supports `Info(args ...interface{})` in their `SugaredLogger`.

fredriv: But, by restricting the API here, we "force" people to put the values in fields instead, which gives us the structured logging we want?

deifyed: Will it be structured unless we force the keys of the fields as well?

Anywho. Valid points all around. I'd use the ...interface{} args version mostly due to having seen it in multiple established places ([go logging lib](https://github.com/golang/go/blob/master/src/log/log.go#L210), logrus, zap). String seems fine enough though. Thoughts @yngvark ?

fredriv: The keys of the fields in `WithField` are strings, if that's what you mean @deifyed? Or do you mean that we standardize names of keys (which could be a good idea anyway)?

deifyed: Last part, standardizing names of keys. I'd argue against standardizing them at this moment since we don't have need for structured logging right now. It was just an argument invalidating the structured part. String is fine for now IMO. 

A use case for using args interface{} is that we could bake in marshalling(serializing) so that we wouldn't have to do this outside the log function everytime we'd want to log a struct for example

deifyed:
Running `make check` will let you see the problems CI is complaining about

pkg/logging/logrus.go
```diff
@@ -0,0 +1,49 @@
...
+		newFields[k] = v
+	}
+	newFields[key] = value
+	return logrusLogger{
+		logger: l.logger,
+		fields: newFields,
+	}
+}
+
+func getLogrusLogger() Logger {
```
yngvark, line 40:
```suggestion
func newLogrusLogger() Logger {
```

It's common in Go to call factory functions for `New` or `NewSomething`.

pkg/logging/types.go
yngvark: > Anywho. Valid points all around. I'd use the ...interface{} args version mostly due to having seen it in multiple established places (go logging lib, logrus, zap).

I agree with this. We're trying to create an interface that abstracts common logging libraries. By not deviating too far from what other loggers do, we're more resilient to future needs for our logging.

Perhaps `WithFields` covers our need for structured logging, but there might be other needs that it doesn't solve. Maybe we want to use a custom logger that always takes some struct as input, for instance.

I don't see any reason to use `string` instead of `interface{}`, other than maybe giving our API clearer intention of use.

On the other hand, we could always change from `string` to `interface{}` in the future without much hazzle, it's the other way around that hurts.

yngvark (commented):
Looks good :+1:

fredriv:
I've updated the PR so it builds cleanly without any linter warnings.

I had to add a `nolint: gochecknoglobals` on the global Logrus instance in `logging/api.go`. Alternatively, if we don't want to use a global logger, we need to inject a logger (via `Okctl`, `Context` or `Config`?) in all the "classes" that we want to log from.

Regarding the `msg string` vs `args ...interface{}` discussion, I would like to postpone changing that until we see an actual need in our codebase for a generic varargs here? (In Logrus, Zap, etc., this variant is only serialized into a log message string, it is not used for structured logging)

Care to take another look @deifyed @yngvark ?

fredriv:
Logging to file is still TODO, but is essentially just a question of adding a Logrus hook, similar to `EnableFileLog` in `config.go`

yngvark (approved):
> I had to add a nolint: gochecknoglobals on the global Logrus instance in logging/api.go. Alternatively, if we don't want to use a global logger, we need to inject a logger (via Okctl, Context or Config?) in all the "classes" that we want to log from.

I think adding the nolint is fine, as in this case that's what we want.

fredriv:
Implemented logging to file as well. Using new logfile `newconsole.log` for the new Logger, to avoid the old and new logger thrashing each other's logs.

pkg/logging/api.go
```diff
@@ -0,0 +1,25 @@
...
+var logger Logger //nolint: gochecknoglobals
+
+// InitLogger initialize the global logger instance
+func InitLogger(logFile string) error {
+	logger, err := newLogrusLogger(logFile)
+	if err != nil {
+		initLogger(logger)
+	}
+
+	return err
```
yngvark, line 14: Shouldn't it be opposite?

```suggestion
	logger, err := newLogrusLogger(logFile)
	if err != nil {
		return err
	}
	
	initLogger(logger)
	
	return nil
```

yngvark (changes requested):
Found a bug, but otherwise good :100:

pkg/logging/api.go
```diff
@@ -0,0 +1,25 @@
+// Package logging API for structured logging
+package logging
+
+// The global logger instance
+var logger Logger //nolint: gochecknoglobals
+
+// InitLogger initialize the global logger instance
+func InitLogger(logFile string) error {
```
deifyed, line 8: Would be nice to not have to initialize the logger. Requiring the logger to be initialized prevents us from logging things that happens before the initialization and I want to be able to log eeeeeverything. I see we're initializing the logger in `o.Initialize`. Some commands doesn't even do `o.Initialize`. Maybe if we move the `InitLogger` call to `cmd/okctl/main.go`

deifyed (changes requested):

pkg/logging/api.go
fredriv: I need some way to initialize it in order to be able to test it. But I can have a go at moving it to `main.go`!

yngvark: We could use the singleton pattern.

fredriv: @deifyed Moved the logger initialization to `main.go`

pkg/logging/api.go
fredriv: Yep! 😅

pkg/logging/api.go
deifyed: nice!

cmd/okctl/main.go
```diff
@@ -47,6 +47,10 @@ func buildRootCommand() *cobra.Command {
 	var outputFormat string
 
 	o := okctl.New()
+	if err := o.InitLogging(); err != nil {
+		fmt.Println("Error configuring logging: %w", err)
```
deifyed, line 51: As far as I know, only fmt.Errorf supports the %w thing(flag?)

Also concider using fmt.Fprintln() instead, with it you can pass output to stderr(os.Stderr)

deifyed (approved):
Nice! Really comming together now

cmd/okctl/main.go
fredriv: 🙈 Fixed, thanks!

pkg/logging/api.go
fredriv: Fixed @yngvark

fredriv:
Even Codecov is happy now 😀

yngvark (approved):
smashing!

PR #827, opened 2021-12-21
KM455 ✅ Add post-mortem template

deifyed:
## Description

Adds an RFC and a template for creating post-mortems.

The template is based on the excellent template made Oslonøkkelen found [here](https://confluence.oslo.kommune.no/pages/viewpage.action?pageId=89102585).

I've translated it and did some minor tweaks, like removing "what did we learn" since I feel like the other fields will adequately convey our learning points. 

I came across a nice source of post-mortem templates [here](https://github.com/dastergon/postmortem-templates).

Remember to view the template in RAW. There's relevant commentary in it. 

Feel free to comment on anything, for example:
- Points that you feel are redundant
- Points that you feel are missing
- Things that should be reworded

The template should be simple to understand and to use. Please review each point critically. Lets make sure we don't write loads of documentation which in hindsight were a waste of time

### DIFF
--- a/docs/post-mortems/0000-post-mortem-template.md
+++ b/docs/post-mortems/0000-post-mortem-template.md
@@ -8,8 +8,11 @@ proaktiv melding.
 -->
 
 ## Timeline
-<!-- 
-Describe the relevant activites in a timeline format. Remember activities leading up to the event being triggered.
+<!--
+Describe the relevant activites in a timeline format. Remember activities
+leading up to the event being triggered. Make sure to link to graphs, logs and
+other relevant information sources.
+
 
 2019-12-30
     23:30 A proactive message got sendt out to all the citizens of Oslo
--- a/docs/post-mortems/0000-post-mortem-template.md
+++ b/docs/post-mortems/0000-post-mortem-template.md
@@ -13,7 +13,6 @@ Describe the relevant activites in a timeline format. Remember activities
 leading up to the event being triggered. Make sure to link to graphs, logs and
 other relevant information sources.
 
-
 2019-12-30
     23:30 A proactive message got sendt out to all the citizens of Oslo
 2019-12-31
@@ -68,12 +67,13 @@ Example:
 - Deploying the database update was fast
 -->
 
-## What could we have done better
+## What went wrong
 <!--
 Describe what went wrong trying to handle the event.
 
 Example:
 - Had to manually roll back the database while testing fixes
+- We missed the notification regarding Keycloak upgrade
 -->
 
 ## Where were we lucky
--- a/docs/post-mortems/0000-post-mortem-template.md
+++ b/docs/post-mortems/0000-post-mortem-template.md
@@ -1,0 +1,109 @@
+## Summary
+<!--
+Describe the reason and the consequences of the event as short and concise as possible.
+
+Example:
+The backend of Oslo Nøkkelen were unavailable for 31 minutes due to an increase in traffic that happened after a
+proaktiv melding.
+-->
+
+## Timeline
+<!-- 
+Describe the relevant activites in a timeline format. Remember activities leading up to the event being triggered.
+
+2019-12-30
+    23:30 A proactive message got sendt out to all the citizens of Oslo
+2019-12-31
+    08:23 Oslonøkkelen had been downloaded 30.000 times, distributed evenly across iOS and Android, during the past 3 hours
+    09:09 DOWNTIME START - backend fails due to the increased load
+    09:14 Received Slack notifications due to an increase in 500 status codes in the backend
+    09:16 Initiated an investigation into the issue
+    09:18 EVENT START - Team creates the Slack channel #citykey-incident-backend to better cooperate on the issue
+    09:20 Manually scaling up the number of backend instances
+    09:24 Found the error. The backend fails when it tries to read an item in the database that doesn't exist. Only happens upon a users first login
+    09:28 Pull request with a fix created: http://github.com/oslokommune/something/something/pr/298
+    09:34 Pull request 298 merged and deployed to production
+    09:39 PROBLEM FIXED - Amount of 500 status code requests are decreasing
+    09:40 DOWNTIME END - Amount of 500 status codes are down to zero
+    10:10 EVENT END - Reaches criteria of 30 minutes with normal activity
+-->
+
+## Impact
+<!--
+Describe the consequences this had for the organization.
+
+Example:
+- Approximately 30.000 users downloaded and experienced an error using the app.
+- Users already logged in experienced some minor delay in the app
+-->
+
+## Cause(s)
+<!--
+Describe relevant factors that played a part in causing the event.
+
+Example:
+The error happened due to a combination of:
+- A spike of new users
+- A bug in the backend code for user creation
+-->
+
+## Solution
+<!--
+Describe how the problem was fixed.
+
+Example:
+- Manually scaled up number of backend instances letting regular traffic stay unaffected
+- Created a fix for the database schema
+-->
+
+## What went well
+<!--
+Describe what went well trying to handle the event.
+
+Example:
+- Alerting mechanisms worked brilliantly when errors started comming in
+- Deploying the database update was fast
+-->
+
+## What could we have done better
+<!--
+Describe what went wrong trying to handle the event.
+
+Example:
+- Had to manually roll back the database while testing fixes
+-->
+
+## Where were we lucky
+<!--
+Describe briefly the situations related to this event where we got lucky.
+
+Example:
+- Updating the database schema fixed the problem
+- Manual horizontal scale up ensured a small amount of users was affected
+-->
+
+## Mitigations
+<!--
+Describe potential steps to prevent this event from happening again, be it technical, processes, etc.
+
+Example:
+- Request that proactive messages will be sendt in smaller batches, ideally during regular work hours, for load to be
+    better distributed.
+- Create tests that cover creation and authentication of users
+-->
+
+## Action items
+<!--
+Describe and reference specific action items
+
+Example:
+[KM123](https://trello.com/c/nJpSSGCs/455-post-mortem-the-github-cert-issue) - Add backoff for login in the iOS app
+[KM221](https://trello.com/c/nJpSSGCs/455-post-mortem-the-github-cert-issue) - Add tests in the backend for creating a user and authenticating
+-->
+
+## Supporting information and documentation
+<!--
+Add relevant additional information or documentation.
+
+Example: images of metrics that show when something went wrong, relevant pieces of logs, etc
+-->
--- a/docs/rfc/0000-rfc-template.md
+++ b/docs/rfc/0000-rfc-template.md
@@ -1,5 +1,4 @@
 # 0000. Request for comments
-
 <!-- 
     Date representing when the RFC was submittet for review 
 
@@ -9,7 +8,6 @@
 Date:
 
 ## Context
-
 <!--
     Short description of why this decision is needed
 
@@ -17,9 +15,7 @@ Date:
     We want to better document our decisions
 -->
 
-
 ## Decision
-
 <!--
     Short summary of the decision
 
@@ -28,11 +24,9 @@ Date:
 -->
 
 ## Consequences
-
 <!--
     Short description of positive and negative consequences of the decision
 
     Example:
     The decisions we deem relevant will be documented in this format here in this folder.
 -->
-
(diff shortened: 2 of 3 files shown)


yngvark (approved):
Looks good :+1:

docs/post-mortems/0000-post-mortem-template.md
```diff
@@ -0,0 +1,109 @@
...
+
+## What could we have done better
+<!--
+Describe what went wrong trying to handle the event.
+
+Example:
+- Had to manually roll back the database while testing fixes
+-->
+
+## Where were we lucky
```
deifyed, line 76: I suggest removing this until we deem it necessary. Unless someone has any arguments for why it should stay?

I trust that everyone in this team will identify luck and make sure it happens by default instead with or without this point

docs/post-mortems/0000-post-mortem-template.md
```diff
@@ -0,0 +1,109 @@
...
+<!--
+Describe briefly the situations related to this event where we got lucky.
+
+Example:
+- Updating the database schema fixed the problem
+- Manual horizontal scale up ensured a small amount of users was affected
+-->
+
+## Mitigations
```
deifyed, line 85: I suggest removing mitigations. I feel it is the same as Action items, just not as specific

docs/post-mortems/0000-post-mortem-template.md
fredriv: I suggest keeping it. We used it actively both in Dataplattform and Schibsted to note things that contributed to the result not being as bad but were outside of our control, such as timing of the incident (e.g. low usage hours).

docs/post-mortems/0000-post-mortem-template.md
fredriv: I suggest keeping it. Here we are supposed to list all the things we can think of to mitigate the issue (brainstorming). Actions are what we decide to actually do (can be a subset, can be things that are not directly mitigations, etc.).

docs/post-mortems/0000-post-mortem-template.md
```diff
@@ -0,0 +1,109 @@
...
+Describe the reason and the consequences of the event as short and concise as possible.
+
+Example:
+The backend of Oslo Nøkkelen were unavailable for 31 minutes due to an increase in traffic that happened after a
+proaktiv melding.
+-->
+
+## Timeline
+<!-- 
+Describe the relevant activites in a timeline format. Remember activities leading up to the event being triggered.
```
fredriv, line 12: Should also include relevant graphs etc.

docs/post-mortems/0000-post-mortem-template.md
fredriv: Also, in incident reviews I've attended, the report is written beforehand by one or a few people, including a list of possible mitigations that they come up with. The action items are only decided upon by the full team during the actual review. (During the review you can also add to/change the mitigations list and other sections of course).

docs/post-mortems/0000-post-mortem-template.md
```diff
@@ -0,0 +1,109 @@
...
+Describe what went well trying to handle the event.
+
+Example:
+- Alerting mechanisms worked brilliantly when errors started comming in
+- Deploying the database update was fast
+-->
+
+## What could we have done better
+<!--
+Describe what went wrong trying to handle the event.
```
fredriv, line 70: This should cover not just the handling of the event, but also issues in our systems, processes, etc. that led to the incident. Similar in the section above. Suggest renaming it to "What went wrong".

Example:
- We did not have any alarms for errors in our service
- We missed the notification about Keycloak being upgraded

fredriv: Can also include things not under our control, e.g. "Internal Enterprise Github was down which hindered us in pushing out a fix" (since we don't have an "Were were we unlucky" section).

fredriv:
Also lots of good incident reports on Confluence: https://confluence.oslo.kommune.no/pages/viewpage.action?pageId=89102154

docs/post-mortems/0000-post-mortem-template.md
fredriv: These examples are all from real incident reports from the Dataplattform team btw :-) (see Confluence link in comments).

docs/post-mortems/0000-post-mortem-template.md
deifyed: Did you get any value from noting down these things?

docs/post-mortems/0000-post-mortem-template.md
deifyed: So "mitigations" is a part of the review process, but "action items" is the value that "mitigations" results in?

If so I'd rather write a comment about it inside action items. I feel the template should only include information that is valueable looking back to. The process itself != the resulting document

fredriv: Action items are what we decide to do. Mitigations is what we could potentially do. Some of those might be relevant as actions later?

fredriv: Action items should not include things that would be "nice to have, but we won't be doing it now". Each line in action items should be a task in Trello.

docs/post-mortems/0000-post-mortem-template.md
fredriv: I think it's useful to note things that worked in our favour, but that we don't necessarily have control over.

Your comment above "everyone in this team will identify luck and make sure it happens by default instead" indicates things that we can control (i.e. so we can make it happen by default)?

fredriv: I see that we have used that section in 4 out of 4 incident reports from Dataplattform on Confluence, so I would say it was useful. You can always leave it empty in your review if it's not relevant. If we see that we don't have a use for the secton, we can always remove it from the template later?

docs/post-mortems/0000-post-mortem-template.md
deifyed: Yeah. Totally agree with action items should be tasks in Trello. I just have a hard time imagining having time enough to skim through old post-mortems looking for tasks that are nice to have :p

docs/post-mortems/0000-post-mortem-template.md
deifyed: Definitely. I'm just used to adding things when we need them, instead of adding things first, then removing them later.

Trying to mitigate wasting time writing things that doesn't get used. Anyways, both approaches work

docs/post-mortems/0000-post-mortem-template.md
fredriv: Typically "Mitigations" will include a lot more than "Action items", and often the action items will be narrower/more specific. I think there is value in seeing the difference, without needing to look at the Git history or including a lot of comments in the "Action items" section that will just clutter it up.

Part of writing the report is writing a list of the potential mitigations you can come up with, so it's not like your'e saving any work by omitting the "Mitigations" section. Instead you are creating more work by needing to decide what should be left as comments, editing items to be more specific actions, etc.

docs/post-mortems/0000-post-mortem-template.md
fredriv: You asked for our $.02. History has shown me that the section is useful 🤷

deifyed: Yeah! And I appriciate it. Just want to test things to make sure we're not doing things "just because we've always done it like this"

fredriv: The incident report is in big part a record of what happened. Being aware of conditions that can work in our favour (or against us) is useful in itself, is it not? Regardless of whether or not it is something we can influence/control and thus take action on.

fredriv: I looked through the incident reports from Oslonøkkelen and Skjemautvikling on Confluence as well, and the majority (but not all) of them also included items in the "We got lucky" seciton.

fredriv: Most of the things that end up in this section don't really belong in "What went well", e.g. "We don't have many active users on the service yet so noone was affected" is not really something that "went well" 😁

deifyed: > The incident report is in big part a record of what happened. Being aware of conditions that can work in our favour (or against us) is useful in itself, is it not?

Might be, I just don't see how. 

> I looked through the incident reports from Oslonøkkelen and Skjemautvikling on Confluence as well, and the majority (but not all) of them also included items in the "We got lucky" seciton.

Included does not equal value. If someone at some point looked it up to answer some kind of question, then I'll agree it has value.

We obviously disagree here. We'll see what the other's chime in post chrismas. At least now they have some pros and cons to work with ^^

fredriv: It has value if it was useful in that incident review itself, to e.g. reason about impact? Otherwise don't include it of course. It cannot be a requirement that it has future value for someone going back to look at the review later?

fredriv:
Thinking about this a bit while making dinner, I wonder if our disagreements above stem from our understanding of the *purpose* of this document? To me, the main purpose of the incident report is to serve as input to the incident review, so the stakeholders can be informed (what happened, why, etc.) and make reasoned decisions about what actions to take (which are also documented in the report). That someone can come along later and learn something useful from the report is nice, but secondary. Thoughts @deifyed? 🤔

deifyed:
> I wonder if our disagreements above stem from our understanding of the purpose of this document?
Makes sense. 

My understanding of the document (or rather what I want the document to be) is to:
1. produce higher quality tasks that we can do to ensure (or reduce the chance) that the event doesn't happen again. 
2. serve as documentation of the event so others can learn from our mistakes and hopefully refrain from doing them. 'Others' here includes us in the event that we forget.

Neither points benefit from writing down things that are outside of our control.

I do however agree that considering things outside our control has value as a means of thoroughly investigating the event, but I argue that they should be left out of the final document. The final document should (in my opinion) be a condensed version with only things relevant for point #2. Less noise increases the probability that people will read it in the future.

deifyed:
Maybe @ivaruf @yngvark would want to chime in on their perceived / desired purpose of the post mortem template

yngvark:
I agree with arguments from both of you, but I'll add my thoughts.

> produce higher quality tasks that we can do to ensure (or reduce the chance) that the event doesn't happen again.

I agree with this, but would like to extend it. If anything from okctl fails, users want to understand the specifics about the failure. Not only to see that the failure doesn't happen again, but to get a deeper understanding of their infrastructure. The post mortem will help them understand to which extent okctl's environment meets their product's requirements (uptime, security, redundancy, etc.).

As to what the post mortem template should contain, this isn't an excact science. I don't think a post mortem is only there in order for someone to take some action, it is also there to build knowledge about the nature of the system impacted.

We need to have enough in the template so the next post mortem author doesn't forget important information. However, the template should just be a list of things to consider to include when writing the post mortem, not stuff that must be present for all post mortems. The authors should be free to skip stuff from the template if they decide it's not important or noise.

docs/post-mortems/0000-post-mortem-template.md
yngvark: I think "Mitigations" can add value, even if nobody scand old post mortems for things to do. It can add useful info about what the post mortem author(s). Just by reading the points, I can get insight into the incident. It doesn't need to result in tasks now or in the future to provide that value.

The background for this opinion is my take on post mortem: I want to learn about important aspects of an incident, just to understand the nature of it. If the post mortem authors can't find any important to say for any point, then they're free to skip it.

docs/post-mortems/0000-post-mortem-template.md
yngvark: I don't have a strong opinion on this, but if a post mortem doesn't have anything to say on this point, the author can just remove it. Also, if we never use it, we can remove it from the template.

fredriv (approved):
LGTM 👍

fredriv:
>     2. serve as documentation of the event so others can learn from our mistakes and hopefully refrain from doing them. [...]
> 
> Neither points benefit from writing down things that are outside of our control.
> 
> I do however agree that considering things outside our control has value as a means of thoroughly investigating the event, but I argue that they should be left out of the final document. The final document should (in my opinion) be a condensed version with only things relevant for point #2. Less noise increases the probability that people will read it in the future.

I think this is the crux of our disagreement. In point 2 you mention learning from our mistakes, but things outside of our control also certainly contribute to how and why the incident occured and how big impact it has. This is important information in order to understand the context of the incident and should be included IMO. But of course, if the section is not relevant for a particular incident review, feel free to skip it :-)

docs/post-mortems/0000-post-mortem-template.md
deifyed: Leaving it

docs/post-mortems/0000-post-mortem-template.md
deifyed: Leaving it

PR #829, opened 2021-12-21
KM455 ✅ Add Github/SSH known_hosts key mismatch post-mortem

deifyed:
## Motivation and Context

17th of november Github removed some SSH keys which caused problems for us


yngvark (approved):
:+1:

PR #848, opened 2022-01-10
Stop install script on error

ivaruf:
Install script will continue even if there is an error

image

By setting the -e flag we prevent this.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 function get_user_agent() {
   if [[ -f ~/.okctl/conf.yml ]]; then
     USER_AGENT=$( (grep -E "userAgent:" ~/.okctl/conf.yml || echo okctl) | sed 's/userAgent://')


yngvark (approved):
The underlying issue seems to be this:
https://github.com/goreleaser/goreleaser/issues/1382

install/install.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
deifyed, line 3: [This](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script) SO post claims `set -e` is bad practice. Probably isn't in this script, but I thought I'd share

deifyed (approved):

install/install.sh
ivaruf: We could (should?) have proper error handling, but it is better to stop the script than to continue and print "Successfully installed ..." at the end, when in  fact it failed.

Was not going to spend a lot of time to adress this, so went for a 'set -e' to at least improve the situation.

PR #853, opened 2022-01-11
KM451: Use direct api instead of rest for Component Service

ivaruf:
Use component core service directly in client api


yngvark (approved):
:+1: :+1:

PR #854, opened 2022-01-11
KM451: Use core service directly in container repository client api

ivaruf:
Use core service directly in container repository client api


yngvark (approved):
:+1: :+1:

PR #864, opened 2022-01-14
KM451: Use paramater core service directly in client

ivaruf:


deifyed (approved):
Ensure `make lint` runs successfully locally. CI is acting up for some reason, I've created a PR to hopefully fix it

PR #870, opened 2022-01-17
KM451: Use core service account service in direct client api

ivaruf:


deifyed (approved):
FYI: I'd like to review the final result before merging it into master. My hypothesis is that it could do with some cleaning after removing most the stuff from server side into client side initialization

PR #879, opened 2022-01-21
KM491 - Give better error message if cluster is not found in state.db

haavardeide:
## Description

If a user have created a cluster, not run the maintenance commands,
then upgraded to 0.0.80 or higher, a "not found" message will be
displayed. This gives a better error message, and will be accompanied
with a okctl.io common-issues for fixing the issue

## Motivation and Context

See [KM491](https://trello.com/c/2YlRCshT/491-after-updating-to-okctl-0080-users-get-error-not-found): let the user know that we cannot find a cluster that hen is trying to apply a command to 

To reproduce: rename `state.db` to ex `state2.db` in your S3 bucket and run `okctl venv -c cluster.yaml`, this will generate a error message about missing cluster

## Additional info

Have added the verify functionality on all commands that download state.db, take a good look at those to ensure there are no side-effects I haven't thought about

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/hooks/state.go
+++ b/cmd/okctl/hooks/state.go
@@ -239,7 +239,8 @@ func VerifyClusterExistsInState(o *okctl.Okctl) RunEer {
 
 		_, err := handlers.Cluster.GetCluster(o.Declaration.Metadata.Name)
 		if err != nil {
-			o.Logger.Debug("Could not find cluster in current state: ", err)
+			log := logging.GetLogger(logComponent, "VerifyClusterExistsInState")
+			log.Debug("Could not find cluster in current state")
 
 			if errors.Is(err, storm.ErrNotFound) {
 				return fmt.Errorf("getting existing cluster %s: %w", o.Declaration.Metadata.Name, err)
--- a/cmd/okctl/hooks/state.go
+++ b/cmd/okctl/hooks/state.go
@@ -230,6 +230,28 @@ func PurgeRemoteState(o *okctl.Okctl) RunEer {
 	}
 }
 
+// VerifyClusterExistsInState ensures we have a cluster in the state before continuing
+// This could happen if maintenance mode was not run after upgrading
+// to 0.0.80 and we don't have a state.db in S3 - see #491
+func VerifyClusterExistsInState(o *okctl.Okctl) RunEer {
+	return func(_ *cobra.Command, _ []string) error {
+		handlers := o.StateHandlers(o.StateNodes())
+
+		_, err := handlers.Cluster.GetCluster(o.Declaration.Metadata.Name)
+		if err != nil {
+			o.Logger.Debug("Could not find cluster in current state: ", err)
+
+			if errors.Is(err, storm.ErrNotFound) {
+				return fmt.Errorf("getting existing cluster %s: %w", o.Declaration.Metadata.Name, err)
+			}
+
+			return fmt.Errorf("getting cluster from state: %w", err)
+		}
+
+		return nil
+	}
+}
+
 func backupStateDatabase(fs *afero.Afero, rootDir string, clusterName string, dbPath string) error {
 	stateBackupPath := path.Join(
 		rootDir,
--- a/cmd/okctl/hooks/state.go
+++ b/cmd/okctl/hooks/state.go
@@ -243,6 +243,14 @@ func VerifyClusterExistsInState(o *okctl.Okctl) RunEer {
 			log.Debug("Could not find cluster in current state")
 
 			if errors.Is(err, storm.ErrNotFound) {
+				_, _ = fmt.Fprintln(
+					o.Out,
+					"Could not find a cluster in your state.db. If you have recently upgraded from okctl version 0.0.80, "+
+						"you need to do some manual maintenance before continuing. See: "+
+						"https://www.okctl.io/common-issues/#i-get-an-error-message-with-getting-existing-cluster-cluster-name-not-found and "+
+						"https://www.okctl.io/concerning-upgrades-and-patches/",
+				)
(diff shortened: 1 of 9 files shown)


cmd/okctl/hooks/state.go
```diff
@@ -230,6 +230,28 @@ func PurgeRemoteState(o *okctl.Okctl) RunEer {
...
+// VerifyClusterExistsInState ensures we have a cluster in the state before continuing
+// This could happen if maintenance mode was not run after upgrading
+// to 0.0.80 and we don't have a state.db in S3 - see #491
+func VerifyClusterExistsInState(o *okctl.Okctl) RunEer {
+	return func(_ *cobra.Command, _ []string) error {
+		handlers := o.StateHandlers(o.StateNodes())
+
+		_, err := handlers.Cluster.GetCluster(o.Declaration.Metadata.Name)
+		if err != nil {
+			o.Logger.Debug("Could not find cluster in current state: ", err)
```
deifyed, line 242: I'd suggest using our new `pkg/logging` package. It's going to be our standard way of logging going forward

Initialising the logger requires the string arguments "component" and "activity", in this case I'd say the component is "hooks" or "stateHook", and the activity is "VerifyClusterExistsInState"

deifyed (approved):
Nice! clean and readable. See my one comment

cmd/okctl/hooks/state.go
```diff
@@ -230,6 +230,29 @@ func PurgeRemoteState(o *okctl.Okctl) RunEer {
...
+// VerifyClusterExistsInState ensures we have a cluster in the state before continuing
+// This could happen if maintenance mode was not run after upgrading
+// to 0.0.80 and we don't have a state.db in S3 - see #491
+func VerifyClusterExistsInState(o *okctl.Okctl) RunEer {
+	return func(_ *cobra.Command, _ []string) error {
+		handlers := o.StateHandlers(o.StateNodes())
+
+		_, err := handlers.Cluster.GetCluster(o.Declaration.Metadata.Name)
+		if err != nil {
+			log := logging.GetLogger(logComponent, "VerifyClusterExistsInState")
```
deifyed, line 242: Not really sure what best practice is here, but I usually initialize all that can be initialized before the return of the RunEer func. I.e between line 236 and 237 here. That's where I'd put everything that doesn't need execution JIT. I'd try moving the state handlers initialization there, and the logger initialization

deifyed (approved):

haavardeide:
https://www.okctl.io/common-issues/#i-get-an-error-message-with-getting-existing-cluster-cluster-name-not-found

docs/release_notes/0.0.86.md
```diff
@@ -7,6 +7,7 @@ KM478 ✅ Add support for authenticating with AWS profiles and AWS SSO (#856)
 ## Bugfixes
 
 KM519 🐛 Upgrade aws-iam-authenticator to support AWS SSO (#874)
+KM491 Better error message if cluster is not found in state.db
```
deifyed, line 10: We usually add the pull request ID as a postfix, as in (#879) 

```suggestion
KM491 👌 Better error message if cluster is not found in state.db (#879) 
```

deifyed (approved):

yngvark (approved):
Looks good :+1:

cmd/okctl/hooks/state.go
```diff
@@ -230,6 +230,29 @@ func PurgeRemoteState(o *okctl.Okctl) RunEer {
...
+			log.Debug("Could not find cluster in current state")
+
+			if errors.Is(err, storm.ErrNotFound) {
+				return fmt.Errorf("getting existing cluster %s: %w", o.Declaration.Metadata.Name, err)
```
yngvark, line 246: To be more specific, I would say `getting state for existing cluster %s: %w`, to emphasize that it's the state that's missing.

cmd/okctl/hooks/state.go
```diff
@@ -230,6 +230,29 @@ func PurgeRemoteState(o *okctl.Okctl) RunEer {
...
+			}
```
yngvark, line 247: From an UX perspective, I think we could give the user more assistance if this happens.

The change we did earlier was a breaking change, and we need the user to do execute some actions in order to fix the cluster. These are the maintenance commands, and we have described them in https://www.okctl.io/concerning-upgrades-and-patches/ .

So I think we could link to that article. For instance:

```golang
if errors.Is(err, storm.ErrNotFound) {
  _, _ = fmt.Fprintln(
    o.Out,
    "The state file for the cluster was not found. If you have recently upgraded from okctl version 0.0.80, "+
      "you need to do some manual maintenance before continuing. See: "+
      "https://www.okctl.io/concerning-upgrades-and-patches/",
  )
  return fmt.Errorf("getting existing cluster %s: %w", o.Declaration.Metadata.Name, err)
}
```

For the longer term, we have a plan to improve UX on most of error messages. Right now they suffer from being purely technical, and offer no solution. In some future, we could write a really user friendly error message, one technical, and offer a solution.

haavardeide: What I did was create a commin issue section with the error message, so that you can search for `getting existing cluster` and get : https://www.okctl.io/common-issues/#i-get-an-error-message-with-getting-existing-cluster-cluster-name-not-found

deifyed: We could. Though I argumented against it due to:
1. We have no structure for delivering user oriented error messages, its a TODO
2. We by principle try to keep the code base focused on the current version, not looking backwards. 

Thats why I thought making it easy to search for the error on [okctl.io](https://okctl.io) was a nice compromise

PR #880, opened 2022-01-21
KM451: Initialize toolchain with extracted core services function

ivaruf:
Cleanup and refactor how okctl, toolchain and clientservices are initialized

### DIFF
--- a/pkg/client/core/api/direct/direct_clients.go
+++ b/pkg/client/core/api/direct/direct_clients.go
@@ -2,8 +2,8 @@ package direct
 
 import "github.com/oslokommune/okctl/pkg/client"
 
-// Clients contains the direct clients
-type Clients struct {
+// ToolChain contains the direct clients
+type ToolChain struct {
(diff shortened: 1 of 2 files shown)


ivaruf:
Doing it this way, we will in have two sets of cores services. This is not really an issue since they are initialized in the same way --- but we probably want to avoid this when refactoring away from direct clients.

pkg/client/core/api/direct/direct_clients.go
```diff
@@ -0,0 +1,22 @@
+package direct
+
+import "github.com/oslokommune/okctl/pkg/client"
+
+// Clients contains the direct clients
+type Clients struct {
```
deifyed, line 6: I suggest calling this struct `Toolchain`. Likely we'll move the actual server side components into this one to remove the abstraction layer later on

deifyed (approved):

pkg/client/core/api/direct/direct_clients.go
ivaruf: :+1:

PR #881, opened 2022-01-24
Km451 remove handlers and rest clients

ivaruf:


cmd/okctl/main.go
```diff
@@ -168,8 +167,6 @@ func buildRootCommand() (*cobra.Command, *okctl.Okctl) {
 			o.Out = cmd.OutOrStdout()
 			o.Err = cmd.OutOrStderr()
 
-			o.SetFormat(core.EncodeResponseType(outputFormat))
```
deifyed, line 171: This doesn't have anything to do with client/server? I don't think we're using it though

pkg/okctl/okctl.go
```diff
@@ -576,11 +543,6 @@ func (o *Okctl) CoreServices() (core.Services, error) {
 		run.NewKubeRun(o.CloudProvider, o.CredentialsProvider.Aws()),
 	)
 
-	awsIamAuth, err := o.BinariesProvider.AwsIamAuthenticator(awsiamauthenticator.Version)
-	if err != nil {
-		return core.Services{}, err
-	}
```
deifyed, line 582: Why remove this?

deifyed (commented):
I'll approve as soon as you reply to the question about removing the auth thing

pkg/okctl/okctl.go
fredriv: It's moved up to line 487?

deifyed: My bad. Shouldn't read big PRs before 0900

deifyed (approved):
Nice!! A major improvement

cmd/okctl/main.go
ivaruf: Format for http requests - that is going away :)

PR #883, opened 2022-01-25
KM522 👌 Don't cache upgrades on disk

yngvark:
## Description

Use temporary storage instead of persistent storage. This will make sure upgrades are never cached, with the benefit that the user will always get the latest binary for upgrades.

## Motivation and Context

[KM522](https://trello.com/c/PVH8XuoG/522-dont-cache-upgrades-on-disk)

Run `okctl upgrade`. While it's running, see that a new dir in `/tmp` has been made. If any uprgades are applicable, they will be put here.

## Types of changes

## Checklist:


deifyed (approved):
LGTM

PR #884, opened 2022-01-25
🐛 Fix linking errors in forward postgres nightly

deifyed:
## Description

Makes the forward postgres test build its own required binary to avoid linking errors

## Motivation and Context

Forward postgres test breaking due to 

```
cmd/nightly/forward-postgres-tester/dist/forward-postgres-tester: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by cmd/nightly/forward-postgres-tester/dist/forward-postgres-tester)
```

## Types of changes

## Checklist:


.github/workflows/nightly/forward-postgres/action.yaml
```diff
@@ -24,6 +24,11 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Build forward-postgres-tester
+      run: |
+        $(cd cmd/nightly/forward-postgres-tester/ && make build)
+      shell: bash
+
```
yngvark, line 31: Maybe pre-optimalization, but it would be possible to add a cache action to avoid building every time.

yngvark (approved):
:+1:

.github/workflows/nightly/forward-postgres/action.yaml
deifyed: Definitely possible. However this small program takes ~10 seconds to build. I'll save it for when we need to optimize yes

PR #888, opened 2022-01-26
KM531 🐛 Fix forwarding stdout and stderr

yngvark:
When running a cmd, we now set cmd.Stdout and Stderr directly, instead of using a scanner to process the cmd's output.

## Description

The best explanation I'm able to do while keeping it short:

In the old implementation suffered from race conditions. The code

```golang
_, err = io.Copy(multiWriter, multiReader)
```

always, but by chance, was able to run when the command had been running for a short while. Thus okctl was able to copy seemingly all output from reader to the writer.

For longer running commands, only the start of the buffer was copied.

Also, for some reason, this made the Run command run forever. Probably because the blocking call `scanner.Scan()` never ended somehow.

## Motivation and Context

[KM531](https://trello.com/b/rimCUYVC/km)

From master (i.e. before using code from this PR):

Run `OKCTL_DEBUG=true okctl upgrade` while making sure that the ArgoCD upgrade will run. It produces tons of output (but you won't see it), making `okctl upgrade` eventually hang.

Then apply this PR, and rerun. You will now see the previosly missing enormous amounts of debug output. The upgrade will succeed.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/binaries/run/run_test.go
+++ b/pkg/binaries/run/run_test.go
@@ -147,17 +147,15 @@ func TestRunCmdOutput(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			var progress bytes.Buffer
+			var out bytes.Buffer
 
-			out, err := tc.run.Run(&progress, []string{tc.output})
+			outResult, err := tc.run.Run(&out, []string{tc.output})
 			assert.NoError(t, err)
 
 			expectedOutput := "wd=working_dir, path=binary_path, env=, args=" + tc.output
 
-			assert.Equal(t, expectedOutput, string(out))
-
-			expectedProgess := []byte(expectedOutput)
-			assert.Equal(t, expectedProgess, progress.Bytes())
+			assert.Equal(t, expectedOutput, string(outResult))
+			assert.Equal(t, expectedOutput, out.String())
 		})
 	}
 }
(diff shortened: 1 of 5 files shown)


pkg/binaries/run/run_test.go
```diff
@@ -214,3 +212,35 @@ func fakeExecCommand(testName string) run.CmdFn {
...
+		assert.NoError(t, err)
+
+		expectedOutput := testStdout + testStderr
+
+		assert.Equal(t, expectedOutput, string(outResult))
+		assert.Equal(t, expectedOutput, out.String())
+	})
+}
+
+func TestRunProcessStdoutAndStderr(t *testing.T) {
```
deifyed, line 237: Can you explain this more? 

I've never seen an `os.Exit()` in a test before.

deifyed: Why does it care about GO_TEST_PROCESS?

and does it really test anything? It just prints out some text?

deifyed (changes requested):
Approved after answering the questions

pkg/binaries/run/run_test.go
yngvark: I'm following the pattern from `TestRunProcessSuccess`. This main use of this function is not to be directly, but by running `go -test.run ...", see https://github.com/oslokommune/okctl/blob/c754d59d85b481da47d31b532a23e76974852d99/pkg/binaries/run/run_test.go#L199

So the flow is:
* The start point is `TestRunCmdStdoutAndStderr`
* In that test, we call `fakeExecCommand("StdoutAndStderr")` which creates a cmd function that executes `go -test.run=TestRunProcessStdoutAndStderr`.
* By calling cmd.Exec, that function is invoked, and it prints stuff to stdout and stderr.
* We collect the output, and do asserts.

GO_TEST_PROCESS is used to ignore the test if run directly when running tests in okctl. GO_TEST_PROCESS is set to 1 in the code flow described above, so we're able to distinguish when the function is called directly, or via the flow above.

Another implementation of all this would be to create a shell file or binary and push it to git (like we do for the upgrade tests, for instance). I did not consider this, I just followed the existing pattern. For code readability, we can consider this.

yngvark: I'm following the pattern from `TestRunProcessSuccess`. This main use of this function is not to be directly, but by running `go -test.run ...", see https://github.com/oslokommune/okctl/blob/c754d59d85b481da47d31b532a23e76974852d99/pkg/binaries/run/run_test.go#L199

So the flow is:
* The start point is `TestRunCmdStdoutAndStderr`
* In that test, we call `fakeExecCommand("StdoutAndStderr")` which creates a cmd function that executes `go -test.run=TestRunProcessStdoutAndStderr`.
* By calling cmd.Exec, that function is invoked, and it prints stuff to stdout and stderr.
* We collect the output, and do asserts.

GO_TEST_PROCESS is used to ignore the test if run directly when running tests in okctl. GO_TEST_PROCESS is set to 1 in the code flow described above, so we're able to distinguish when the function is called directly, or via the flow above.

Another implementation of all this would be to create a shell file or binary and push it to git (like we do for the upgrade tests, for instance). I did not consider this, I just followed the existing pattern. For code readability, we can consider this.

yngvark: If there is no `os.Exit(0)`, the test fails with this:

```
        	            	expected: "this is stdout\nthis is stderr\n"
        	            	actual  : "this is stdout\nthis is stderr\nPASS\n"
```

I have no idea what the "PASS" is.

deifyed: Ahhh right. That explains it. Thanks

deifyed (approved):

PR #899, opened 2022-02-01
KM266 validate postgres database name

haavardeide:
## Description

The following restrictions apply to a RDS postgres database name:
* Two reserved names that cannot be used as dbName: "db" and "database"
* 1 to 60 alphanumeric characters or hyphens
* First character must be a letter
* Can't end with a hyphen.
* Can't contain two consecutive hyphens
* Don't allow the use of uppercase letters in databases.postgres.name - RDS will lowercase it anyway, and the cloudformation script creating S3 bucket will fail

## Motivation and Context

Stop the execution of reconciler as early as possible to avoid waiting until cloudformation reports back an error.

Set `database.postgres.name` to:
* `db`
* `database`
* `usersusersusersusersusersusersusersusersusersusersusersusers1`
* `9-nine`
* `nine-`
* `nine--nine`
to see a error from the postgres reconciler

Set `database.postgres.name` to `your-database-name` to verify that the validation doesn't hinder creation of a fresh database

## Additional info

Validation is only done on create

## Types of changes

## Checklist:


pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -303,15 +306,50 @@ type ClusterDatabasesPostgres struct {
 
 // Validate the content of a postgres database declaration
 func (c ClusterDatabasesPostgres) Validate() error {
+	const minLengthDatabaseName int = 1
+
+	const maxLengthDatabaseName int = 60
```
deifyed, line 311: I'd move these to the const declaration on line 22. No need to declare these each time Validate() is called

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -303,15 +306,50 @@ type ClusterDatabasesPostgres struct {
...
+
+	/*
+	 * Validation of database name is based on:
+	 * 1. AWS RDS rules when creating a postgres database
```
deifyed, line 315: Would love a link to AWS documentation where they describe the rules

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
```diff
@@ -303,15 +306,50 @@ type ClusterDatabasesPostgres struct {
...
+		),
+		validation.Field(&c.User, validation.Required, validation.NotIn(
 			"admin",
-		})),
+		).Error("'admin' is a reserved postgres username")),
 		validation.Field(&c.Namespace, validation.Required),
 	)
 }
 
+func fieldCanNotContainString(str string, errorString string) validation.RuleFunc {
```
deifyed, line 340: I think the file readability could benifit from moving this function to it's own file in the current directory. Maybe `custom-validators.go`

deifyed (approved):
Nice!!

pkg/apis/okctl.io/v1alpha1/cluster_v1alpha1.go
haavardeide: well, there is always https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CreateDBInstance.html - but that doesn't explain what the RDS interface actually does (there is more information/restrictions described in the RDS console)

deifyed: Oh right. I thought they'd have a nice name validation list

PR #901, opened 2022-02-07
👌 Use AWS-managed policy for Lambda VPC access

fredriv:
Replace local policy definition with AWS-managed policy for giving RDS password rotator Lambda function the necessary permissions to run in VPC.

## Description

Internal refactoring, no changes in behavior.

## Motivation and Context

Use AWS-managed policy instead of maintaining our own.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/cfn/components/composer.go
+++ b/pkg/cfn/components/composer.go
@@ -1348,6 +1348,7 @@ const (
 	// - https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Monitoring.OS.html
 	amazonRDSEnhancedMonitoringRole = "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole"
 	awsLambdaBasicExecutionRole     = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+	awsLambdaVPCAccessExecutionRole = "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
 )
 
 // CloudFormationResourceName returns the CF logical name
@@ -1471,7 +1472,7 @@ func (c *RDSPostgresComposer) Compose() (*cfn.Composition, error) {
 		c.PhysicalName("RDSPGRotaterRole"),
 		c.CloudFormationResourceName("RDSPGRotaterRole"),
 		v1alpha1.PermissionsBoundaryARN(c.AWSAccountID),
-		[]string{awsLambdaBasicExecutionRole},
+		[]string{awsLambdaBasicExecutionRole, awsLambdaVPCAccessExecutionRole},
 		policydocument.PolicyDocument{
 			Version: policydocument.Version,
 			Statement: []policydocument.StatementEntry{
@@ -1494,17 +1495,6 @@ func (c *RDSPostgresComposer) Compose() (*cfn.Composition, error) {
 						Effect: policydocument.EffectTypeAllow,
 						Action: []string{
 							"secretsmanager:GetRandomPassword",
-						},
-						Resource: []string{
-							"*",
-						},
-					},
-					{
-						Effect: policydocument.EffectTypeAllow,
-						Action: []string{
-							"ec2:CreateNetworkInterface",
-							"ec2:DeleteNetworkInterface",
-							"ec2:DescribeNetworkInterfaces",
 						},
 						Resource: []string{
 							"*",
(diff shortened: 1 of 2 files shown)


pkg/cfn/components/composer.go
```diff
@@ -1348,6 +1348,7 @@ const (
 	// - https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Monitoring.OS.html
 	amazonRDSEnhancedMonitoringRole = "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole"
 	awsLambdaBasicExecutionRole     = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+	awsLambdaVPCAccessExecutionRole = "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
```
deifyed, line 1351: This role, does it exist in all clusters by default?

yngvark:
Do we want to update existing clusters? In that case, we need an upgrade for this.

Strictly speaking, whenever we change the recipe for how an Okctl cluster is created, we need to update existing ones, to avoid drift.

pkg/cfn/components/composer.go
fredriv: Yes, all the AWS managed policies are available everywhere.

deifyed (approved):
Nice!

fredriv:
> Do we want to update existing clusters? In that case, we need an upgrade for this.
> 
> Strictly speaking, whenever we change the recipe for how an Okctl cluster is created, we need to update existing ones, to avoid drift.

Hmm, good question. What would this entail, modifying the existing `okctl-rdspostgres-*` CloudFormation stack using a change set? Is that possible without taking down the entire stack (and thus the RDS database)?

This doesn't really change any permissions, just makes the role more future-proof (and fewer policies maintained by us). Is it necessary / worth it?

deifyed:
> Hmm, good question. What would this entail, modifying the existing okctl-rdspostgres-* CloudFormation stack using a change set? Is that possible without taking down the entire stack (and thus the RDS database)?
I've patched the CFN stack earlier by downloading the existing stack, patching it, then uploading the updated stack. I doubt it will be recreated, but definitely worth testing. 

Changesets might be the better way, but I've yet to research two questions I have regarding them:
1. Is there a ceiling for how many changesets we can have? If so, we shouldn't use them unless theres a way to reset that number later on. We can't put a ceiling on how many upgrades we can do
2. I've read somewhere that you can't really know if CFN will recreate resources when updating them through changesets. Is that true?

> This doesn't really change any permissions, just makes the role more future-proof (and fewer policies maintained by us). Is it necessary / worth it?

I'm leaning towards that its worth it. All future okctl versions will assume that we are using a managed role, which means any changes and/or upgrades might break due to this assumption. I'd rather not have to remember this detail and make logic for handling both cases down the line. We've also designed upgrades to be very bare bone, meaning you're free to use any way you find reasonable to do this. In other words, no potentially complicated codebase to learn. It can basically be a script

fredriv:
> I've patched the CFN stack earlier by downloading the existing stack, patching it, then uploading the updated stack. I doubt it will be recreated, but definitely worth testing.
> 
> Changesets might be the better way, but I've yet to research two questions I have regarding them:
> 
>     1. Is there a ceiling for how many changesets we can have? If so, we shouldn't use them unless theres a way to reset that number later on. We can't put a ceiling on how many upgrades we can do

There is a (soft) limit of 2000 CF stacks, but I could not find anything specifically about change sets: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

>     2. I've read somewhere that you can't really know if CFN will recreate resources when updating them through changesets. Is that true?

According to [the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-changesets.html) CF will show what resources are replaced, but it will not necessarily be able to tell you whether a change set will succeed (because of resource limits, permissions, etc.).

> I'm leaning towards that its worth it. All future okctl versions will assume that we are using a managed role, which means any changes and/or upgrades might break due to this assumption. I'd rather not have to remember this detail and make logic for handling both cases down the line. We've also designed upgrades to be very bare bone, meaning you're free to use any way you find reasonable to do this. In other words, no potentially complicated codebase to learn. It can basically be a script

I can investigate creating a change set, but since this will need to be applied to all RDS stacks I guess we need to at least read the state DB in order to know which ones to modify.

PR #903, opened 2022-02-09
Km451 remove GitHub direct

ivaruf:
Remove last direct client and direct package.


deifyed (approved):
Nice! This is looking much more natural imo :D

yngvark (approved):

PR #907, opened 2022-02-14
KM321 - 👌 ensure database name is valid as early as possible before setting up postgres forward

haavardeide:
## Description

Currently you would not be notified before pgbouncer is set up and
trying to connect to the datbase. All databases are available in
the state: we now check against one of these before setting up the
pgbouncer

## Motivation and Context

When validating the database as early as possible we save time and frustration for the user.

Run 
```
okctl forward postgres --cluster-declaration cluster.yaml --name notexistingdatabase --username bob --password-file secret/tmp.password
```
and get a error stating that the name is not valid. In addition a list of available databases will be printed to the user.

## Types of changes

## Checklist:


yngvark (approved):
Nice to finally get this validation :+1:

PR #908, opened 2022-02-14
KM483 🐛 Write kubeconfig on okctl upgrade

yngvark:
Write kubeconfig when running `okctl upgrade`.

## Description

Also removed duplication using a WriteKubeConfig hook.

## Motivation and Context

https://trello.com/c/W2rTY7In/483-write-kubeconfig-before-running-upgrades

*  `rm ~/.okctl/credentials/clusterName/kubeconfig`
* Run `okctl upgrade`
* Verify that there is no error
* Verifiy that `ls ~/.okctl/credentials/clusterName/kubeconfig` now returns a file

## Types of changes

## Checklist:

### DIFF
--- a/cmd/okctl/hooks/kubeconfig.go
+++ b/cmd/okctl/hooks/kubeconfig.go
@@ -15,6 +15,7 @@ func WriteKubeConfig(o *okctl.Okctl) RunEer {
 			return fmt.Errorf("getting kubeconfig store: %w", err)
 		}
 
+		// GetKubeConfig actually writes the config to disk
 		_, err = kubeconfigStore.GetKubeConfig(o.Declaration.Metadata.Name)
 		if err != nil {
 			return fmt.Errorf("getting kubeconfig: %w", err)
--- a/cmd/okctl/hooks/kubeconfig.go
+++ b/cmd/okctl/hooks/kubeconfig.go
@@ -1,0 +1,25 @@
+package hooks
+
+import (
+	"fmt"
+
+	"github.com/oslokommune/okctl/pkg/okctl"
+	"github.com/spf13/cobra"
+)
+
+// WriteKubeConfig writes the Kubernetes configuration to disk
+func WriteKubeConfig(o *okctl.Okctl) RunEer {
+	return func(_ *cobra.Command, _ []string) error {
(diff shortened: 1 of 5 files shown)


cmd/okctl/hooks/kubeconfig.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+
+	"github.com/oslokommune/okctl/pkg/okctl"
+	"github.com/spf13/cobra"
+)
+
+// WriteKubeConfig writes the Kubernetes configuration to disk
+func WriteKubeConfig(o *okctl.Okctl) RunEer {
```
deifyed, line 11:
```suggestion
func EnsureKubeConfig(o *okctl.Okctl) RunEer {
```
Maybe?

cmd/okctl/venv.go
```diff
@@ -64,40 +62,6 @@ func buildVenvCommand(o *okctl.Okctl) *cobra.Command { //nolint: funlen
...
-			err = o.FileSystem.MkdirAll(path.Dir(kubeConfigFile), 0o700)
-			if err != nil {
-				return err
-			}
-
-			err = o.FileSystem.WriteFile(kubeConfigFile, data, 0o640)
-			if err != nil {
-				return err
-			}
-
```
deifyed, line 100: NICE :o

cmd/okctl/hooks/kubeconfig.go
yngvark: Hmm, naming :smile: 

I used WriteKubeConfig because that's exactly what it does. I could and perhaps even have been more specific, WriteKubeConfigToDisk.

EnsureKubeConfig is more vague, which is not bad, but for me it communicates something else, namely that "make sure the Kubeconfig exists in whatever way we need, but we don't care about which way".

When it comes to Kubeconfig, this is a file that must exist on disk, so I'm still leaning against `WriteKubeConfig`.

deifyed: Sure! I was thinking from the perspective of the user (developer, command); as a command, I want to make sure the kubeconfig exists. Doesn't really matter. You do you

PR #910, opened 2022-02-15
KM303 Add delete application

deifyed:
## Description

- Extracts application handling into a function so it can be used by both `apply application` and `delete application`
- Adds empty ECR repository functionality required for deleting populated ECR repositories
- Adds delete ArgoCD manifest (locally and remotely) functionality
- Adds delete postgres integration functionality
- Adds functionality for checking existence and deleting resources through kubectl
- Extracts SSL certificate into its own reconciler
- Adds mock default logger to avoid nil pointer exceptions when missing initialization

## Motivation and Context

[KM303](https://trello.com/c/aTG8Bdgr)

1. Create an application with N (`scaffold application` and `apply application`) integrations (ingress, postgres, ECR). 
2. Push to git remote and watch it come alive
3. Run `okctl delete application -f app.yaml` to watch it become less alive

## Types of changes

## Checklist:


cmd/okctl/handlers/application.go
```diff
@@ -0,0 +1,179 @@
...
+
+// HandleApplicationOpts contains all the necessary options for application reconciliation
+type HandleApplicationOpts struct {
+	Okctl *okctl.Okctl
+
+	File string
+
+	Application v1alpha1.Application
+	Purge       bool
+	Confirm     bool
```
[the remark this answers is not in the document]
deifyed: Yes! Nice catch. I totally forgot :D

pkg/api/core/cloudprovider/aws/container_repository_aws.go
```diff
@@ -55,6 +59,25 @@ func (c *containerRepositoryCloudProvider) DeleteContainerRepository(opts *api.D
...
+	if err != nil {
+		return fmt.Errorf("acquiring image IDs: %w", err)
+	}
+
+	_, err = c.provider.ECR().BatchDeleteImageWithContext(ctx, &ecr.BatchDeleteImageInput{
+		ImageIds:       result.ImageIds,
+		RepositoryName: aws.String(opts.Name),
+	})
+	if err != nil {
+		return fmt.Errorf("deleting images: %w", err)
```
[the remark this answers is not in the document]
deifyed: Sounds like maybe the app didn't have any images in its repository? Maybe I forgot to ensure idempotency

yngvark (approved):
Changes LGTM :+1:

cmd/okctl/handlers/application.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	return commands.WriteApplyApplicationSuccessMessage(commands.WriteApplyApplicationSucessMessageOpts{
+		Out:         out,
+		Cluster:     cluster,
+		Application: application,
+	})
+}
+
+const deleteApplicationPromptTemplate = `
```
haavardeide, line 107: ➕ 
Good overview of actually resources being deleted when applying this command!

cmd/okctl/handlers/application_test.go
```diff
@@ -0,0 +1,55 @@
+package handlers
```
haavardeide, line 1: No tests to ensure that the `apply application` path through HandleApplication will execute correctly. The `opts.Purge` and `preConfirmed` variables are important for correct behavior

cmd/okctl/handlers/application.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		state := opts.Okctl.StateHandlers(opts.Okctl.StateNodes())
+
+		services, err := opts.Okctl.ClientServices(state)
+		if err != nil {
+			return fmt.Errorf("acquiring client services: %w", err)
+		}
+
+		kubectlClient := binary.New(
```
haavardeide, line 48: line 48 to 80: to better the readability it could have been something like: `scheduler := createScheduler(opts)` - then the boilerplate of setting up the scheduler doesn't interfere with the execution of the handler.

pkg/client/core/service_application_client.go
```diff
@@ -127,6 +113,165 @@ func (s *applicationService) CreateArgoCDApplicationManifest(opts client.CreateA
...
+			return false, nil
+		}
+
+		return false, errors.E(err, "checking existence of ArgoCD application manifest: %w", err)
+	}
+
+	return true, nil
+}
+
+// deleteFileInRemote creates a commit on the main branch of a repository that removes a file path
```
haavardeide, line 183: not only creates, but also pushes the commit

pkg/client/core/service_application_client.go
```diff
@@ -127,6 +113,165 @@ func (s *applicationService) CreateArgoCDApplicationManifest(opts client.CreateA
...
 }
 
+// DeleteArgoCDApplicationManifest removes necessary files related to the ArgoCD integration
+// This function adds a finalizer to the ArgoCD application manifest before deletion, ensuring a cascading delete.
+// https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/#about-the-deletion-finalizer
+func (s *applicationService) DeleteArgoCDApplicationManifest(opts client.DeleteArgoCDApplicationManifestOpts) error {
+	relativeArgoCDApplicationManifestPath := getRelativeArgoCDManifestPath(opts.Cluster, opts.Application)
+	absoluteArgoCDApplicationManifestPath := path.Join(s.absoluteRepositoryDir, relativeArgoCDApplicationManifestPath)
+
+	err := s.deleteFileInRemote(opts.Cluster, opts.Application, relativeArgoCDApplicationManifestPath)
```
haavardeide, line 123: Reading this function name doesn't give the full impression of what it does. `deleteManifestFromGitRepository()`?

pkg/client/core/service_application_client.go
```diff
@@ -127,6 +113,165 @@ func (s *applicationService) CreateArgoCDApplicationManifest(opts client.CreateA
...
+	if err != nil {
+		return fmt.Errorf("deleting application manifest in repository: %w", err)
+	}
+
+	err = s.fs.Remove(absoluteArgoCDApplicationManifestPath)
+	if err != nil {
+		if stderrors.Is(err, os.ErrNotExist) {
+			return nil
```
haavardeide, line 131: If I read this correctly: if we try to remove a manifest, and it doesn't exist, we just return nil? Is this in case we run this operation several times, and we have already removed it?

I see we have several of this `nil` return cases in this patch, so I guess it is all about having done it before. But what if it is the wrong input to this function? Or something changes, wouldn't this just silence any error?

pkg/controller/application/reconciliation/application_reconciler.go
```diff
@@ -107,28 +144,93 @@ func (a *applicationReconciler) hasCreateDependenciesMet(meta reconciliation.Met
...
+
+	exists, err := state.ContainerRepository.ApplicationHasImage(application.Metadata.Name)
+	if err != nil {
+		return false, fmt.Errorf("determining existence of a ECR repository: %w", err)
+	}
+
+	return exists, nil
+}
+
+func hasIngressCreateDependencies(state *clientCore.StateHandlers, application v1alpha1.Application) (bool, error) {
```
haavardeide, line 181: follow the same naming as the rest of these functions: `hasIngressCreateDependenciesMet()`?

haavardeide (approved):
Big patch :)
Some comments regarding naming and unclear practice, but other than that looks good

cmd/okctl/handlers/application.go
deifyed: 💪

pkg/client/core/service_application_client.go
deifyed: turned it into "deleteFileFromGitRepository"

pkg/client/core/service_application_client.go
deifyed: Hmm good point. It's supposed to ensure that all parts get handled if something crashes mid action. I see theres faults in the logic writes code

pkg/client/core/service_application_client.go
deifyed: You can't create a commit on a remote branch without actually pushing it :D but I'll rephrase

cmd/okctl/handlers/application_test.go
deifyed: I was unable to test preConfirmed due to our prompt lib having a less than ideal way of configuring IO, but purge flag tests are added

deifyed: @haavardeide

PR #930, opened 2022-03-04
KM260 👌 - use '-f' and not '-c' for cluster reference

haavardeide:
## Description

Use `-f` instead of `-c` for cluster file when running `okctl delete cluster`

## Motivation and Context

Delete is not in the context of a cluster (-c) 

* Running `okctl delete cluster -c cluster.yaml`  should result in a error
* Running `okctl delete cluster -f cluster.yaml`  should result in a delete operation 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


deifyed (changes requested):
💪💪💪. This should probably be announced since it'l break any systems relying on the API. Speaking of. This PR should include a change to the nightly delete cluster action

haavardeide:
> 💪💪💪. This should probably be announced since it'l break any systems relying on the API. Speaking of. This PR should include a change to the nightly delete cluster action

Yes, we need to announce on support channel!

haavardeide:
Code analysis issue should be solved in a separate issue: https://trello.com/c/vkuUElKa/619-rework-delete-create-cluster-to-use-handler-pattern

deifyed (approved):
Fire!!

yngvark:
@haavardeide The delete command now uses `--file`, but other commands still use `--cluster-declaration`. Do we have a plan or thoughts about how to deal with this inconcistency?

haavardeide:
> @haavardeide The delete command now uses `--file`, but other commands still use `--cluster-declaration`. Do we have a plan or thoughts about how to deal with this inconcistency?

The plan was not to get rid of `--cluster-declaration`, but to use `--file` in the delete cluster context.

yngvark:
Ah, scratch my comment, there is no inconcistency. I just realized okctl apply cluster always have used `--file`.

yngvark:
@haavardeide The release notes went into 0.0.93, but it should have been insterted into 0.0.94. Can you update please?

PR #932, opened 2022-03-10
KM592 👌 - Use ozzo to validate github/aws credential types from user

haavardeide:
## Description
Use custom types for AWS/Github credential types to ensure correct input from the user

## Motivation and Context

Follow the same pattern as other input values to ensure a good way of validating input values from the user as early as possible

Run commands with `-g` or `-a` with both correct and invalid values to see that the `Validate()` function gives a good feedback to the user on which values are accepted

## Additional info

We are still using global values to reference the input from the user. We have a ticket on removing the global, but for now we keep the global to make it easy to use `addAuthenticationFlags()` without making it too complex

## Types of changes

## Checklist:


cmd/okctl/auth/credentials.go
```diff
@@ -1,47 +1,90 @@
...
 )
 
-var (
-	// AwsCredentialsType is the chosen AWS credential type
-	AwsCredentialsType string //nolint:gochecknoglobals
-	// GithubCredentialsType is the chosen Github credential type
-	GithubCredentialsType string //nolint:gochecknoglobals
-)
+// AwsCredentialsType is the chosen AWS credential type
+// Should only be accessed by main.addAuthenticationFlags()
```
yngvark, line 18: Nice comment

cmd/okctl/main.go
```diff
@@ -78,7 +78,7 @@ func addAvailableCommands(cmd *cobra.Command, o *okctl.Okctl) {
 // Add the common authentication flags used throughout the application.
 // Each sub-command needs to apply this individual according to needs
 func addAuthenticationFlags(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVarP(&auth.AwsCredentialsType,
+	cmd.PersistentFlags().StringVarP((*string)(&auth.AwsCredentialsType),
```
yngvark, line 81: I cannot remember seeing this (*string) syntax before. TIL :100:

yngvark (approved):
:+1:

PR #988, opened 2022-08-15
22Q2-18 🐛 Fix delete app not regarding env

deifyed:
## Description

- Deletes overlay directory for the relevant cluster in all cases
- Introduces a check to see if there is 0 associated clusters, which if true,
    we delete the whole application root directory

## Motivation and Context

[22Q2-18](https://trello.com/c/0ZBgCAbZ)

### Test the case of no related clusters
1. Add a new application with `okctl apply application`
2. Run `okctl delete application` and verify that the application root
    directory is deleted.

### Test the case of one or more related clusters
1. Add a new application with `okctl apply application`.
2. Run `cp -r infrastructure/<yourcluster>{,-mock}` and
    `cp -r infrastructure/applications/<app>/overlays/<yourcluster>{,-mock}`
	to emulate another associated cluster.
3. Run `okctl delete application` and verify that only the relevant overlays
    directory for the relevant cluster is deleted and that the application
	root directory is still intact.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/client/core/service_application_client.go
+++ b/pkg/client/core/service_application_client.go
@@ -6,10 +6,15 @@ import (
 	"encoding/json"
 	stderrors "errors"
 	"fmt"
+	fsPkg "io/fs"
 	"os"
 	"path"
+	"strings"
 
+	"github.com/oslokommune/okctl/pkg/paths"
 	"github.com/oslokommune/okctl/pkg/scaffold/resources"
+
+	"sigs.k8s.io/yaml"
 
 	"github.com/oslokommune/okctl/pkg/clients/kubectl"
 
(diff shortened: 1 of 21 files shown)


pkg/client/core/service_application_client.go
```diff
@@ -86,14 +87,120 @@ func (s *applicationService) ScaffoldApplication(ctx context.Context, opts *clie
...
+		if valid {
+			applications = append(applications, strings.Replace(item.Name(), path.Ext(item.Name()), "", 1))
+		}
+	}
+
+	return applications, nil
+}
+
+// AmountAssociatedClusters knows how to count the number of associated clusters a specific application has. A cluster is
+// classified as associated iff the cluster has an ArgoCD application referencing the app and the app has an overlay folder
```
yngvark, line 148:
```suggestion
// classified as associated if the cluster has an ArgoCD application referencing the app and the app has an overlay folder
```

pkg/client/core/service_application_client.go
```diff
@@ -86,14 +87,120 @@ func (s *applicationService) ScaffoldApplication(ctx context.Context, opts *clie
...
+// referencing the cluster.
+func AmountAssociatedClusters(fs *afero.Afero, absoluteRepositoryRootDir string, clusterContext v1alpha1.Cluster, app v1alpha1.Application) (int, error) {
```
yngvark, line 150: Consider: A common way I've seen of naming functions and variables is using the word "count", for instance as a noun
* `GetAssoicatedClustersCount`

or a verb:
* `CountAssociatedClusters`

pkg/client/core/service_application_client.go
```diff
@@ -86,14 +87,120 @@ func (s *applicationService) ScaffoldApplication(ctx context.Context, opts *clie
...
+	associatedClusters, err := AmountAssociatedClusters(s.fs, s.absoluteRepositoryDir, opts.Cluster, opts.Application)
+	if err != nil {
+		return fmt.Errorf("counting associated clusters: %w", err)
+	}
+
+	if associatedClusters != 0 {
+		return nil
+	}
+
+	absoluteApplicationRootDirectory := path.Join(
```
yngvark, line 198: I find I have to go through a lot of variables and function calls when I'm trying to figure out what this variable really contains. Could we perhaps add a comment that gives an example of what this variable contains? For instance:

```go
// Example: /home/me/my-iac-repo/infrastructure/applications/my-app
```

If so, consider above line 179 as well.

One counter argument might be that it we don't do it other places in the code.

pkg/client/core/service_application_client.go
```diff
@@ -36,6 +36,7 @@ type applicationService struct {
 	fs                    *afero.Afero
 	absoluteRepositoryDir string
 	kubectl               kubectl.Client
+	gitDeleteRemoteFileFn client.GitDeleteRemoteFileFn
```
yngvark, line 39: Consider using an interface instead of passing a function. Using interfaces enables the code reader to simply press CTRL+ALT+B (in intellij) to find the possible implementations. It might not be needed for simple cases, but here we have to backtrack via the New function in service_application_client.go in line 399 before we're able to find out the actual contents of the function - which increases mental load and number of operations in order to find the implementation.

yngvark (approved):
Good testing :+1:

pkg/client/core/service_application_client.go
deifyed: "Iff" is nerd for "if and only if", I'll write the whole thing since it wasn't obvious

deifyed:
@yngvark I've updated the code. Since you approved I'll merge when tests are green. Check out the changes if you want

PR #990, opened 2022-08-17
22Q2-44 ✅ Replace networkingv1beta1 with networkingv1 for Ingress

yngvark:
## Description

Replace ingress version networkingv1beta1 with networkingv1.

## Motivation and Context

[22Q2-44](https://trello.com/c/POlaUKOa/44-bytte-fra-ingress-ingress-networkingv1beta1-til-v1)

* Run `okctl apply application` on an existing or a new application.
* Inspect `ingress.yaml` in the application's directory.
* The version should be networkingv1.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/scaffold/resources/ingress.go
+++ b/pkg/scaffold/resources/ingress.go
@@ -4,16 +4,15 @@ import (
 	"fmt"
 
 	"github.com/oslokommune/okctl/pkg/apis/okctl.io/v1alpha1"
-	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // CreateOkctlIngress creates an ingress customized for okctl
-func CreateOkctlIngress(app v1alpha1.Application) (networkingv1beta1.Ingress, error) {
+func CreateOkctlIngress(app v1alpha1.Application) (networkingv1.Ingress, error) {
 	ingress, err := createGenericIngress(app)
 	if err != nil {
-		return networkingv1beta1.Ingress{}, err
+		return networkingv1.Ingress{}, err
 	}
 
 	ingress.Spec.Rules[0].HTTP.Paths[0].Path = "/*"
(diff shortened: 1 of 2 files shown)


pkg/scaffold/resources/ingress.go
```diff
@@ -24,61 +23,77 @@ func CreateOkctlIngress(app v1alpha1.Application) (networkingv1beta1.Ingress, er
...
 	}
 
 	ingress := generateDefaultIngress()
 	ingress.ObjectMeta.Namespace = app.Metadata.Namespace
 
 	ingress.ObjectMeta.Name = app.Metadata.Name
 
-	ingress.Spec.Rules = []networkingv1beta1.IngressRule{
+	// https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/tasks/ssl_redirect/
+	pathType := networkingv1.PathTypeImplementationSpecific
```
deifyed, line 79: Are you sure `ImplementationSpecific` path type suits non-https-redirect routes?

ref: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types

I'd assume `Prefix` was the most natural choice

deifyed (approved):
Approved when my comment regarding PathType is answered. 

We should also remember that while our apps might support ingressv1, some of our packaged components does not. An example is kubepromstack

pkg/scaffold/resources/ingress.go
yngvark: To be honest, I am not sure, but I am going to test it. I first used Prefix, but then I saw the AWS specific Ingress example using PathTypeImplementationSpecific (see doc in line 78).

Although that example use

```
 alb.ingress.kubernetes.io/ssl-redirect
```

while we use

```
alb.ingress.kubernetes.io/actions.ssl-redirect
```

so I need to dig a bit more into this.

Also, the link you provided says

> Implementations can treat this as a separate pathType or treat it identically to Prefix or Exact path types.

which _can_ mean that the ALB ingress controller can choose to treat it like Prefix, so I was thinking the ALB ingress controller knows best. But yeah, I'll more into it.

deifyed: I'm sure implementation specific is correct for the ssl redirect route, I just assumed prefix was the natural one for when you didn't want to handle routing somewhere else. Super interested in the result of your research

yngvark:
Closing pull request as we won't be working with this any time soon.

PR #1004, opened 2022-08-30
22Q2-55 ✅  Use TLS 1.2 only in ArgoCD and Grafana

yngvark:
## Description

Disables TLS 1.0 and 1.1 in ArgoCD and Grafana, leaving 1.2 to be used.

## Motivation and Context

[22Q2-55](https://trello.com/c/MbM4lznT/55-fjerne-tls-10-and-11-for-alb-er)

Fire up a new cluster. Verify that endpoints for argocd and grafana only supports TLS 1.2. See steps here for sslscan to figure out how: #1003 

## Types of changes

## Checklist:


yngvark:
Won't fix as we're not taking the time to create an upgrade for it. This security improvement is not prioritized as most users are using modern, updated browsers which will prefer TLS 1.2 anyway. We can implement this later if we want it.

